PR #43, opened 2020-05-03
Fix #40: allow the Operator to start on Openshift 3.x

### LINKED ISSUE #40: The Operator fails to start on Openshift 3.x
When starting the Operator on Openshift 3.x the following error occurs:

~~~
2020-05-02T19:57:51.816-0300	ERROR	cmd	Manager exited non-zero	{"error": "no matches for kind \"Ingress\" in version \"networking.k8s.io/v1beta1\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/lcaparel/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
main.main
	/home/lcaparel/gitrepos/nexus-operator/cmd/manager/main.go:155
runtime.main
	/usr/local/go/src/runtime/proc.go:203
FATA[0012] Failed to run operator locally: failed to run operator locally: failed to exec []string{"build/_output/bin/nexus-operator-local"}: exit status 1 
~~~

This is due to the fact that we only compare group names when assessing which resources are available in `pkg/framework/controller_watcher.go`:

~~~
for _, object := range watchedObjects {
	// core resources
	if object.AddToScheme == nil {
		desiredObjects = append(desiredObjects, object)
	} else {
		found := false
		for _, serverGroup := range serverGroups.Groups {
			if strings.Contains(serverGroup.Name, object.GroupVersion.Group) {
				addToScheme = append(addToScheme, object.AddToScheme)
				desiredObjects = append(desiredObjects, object)
				found = true
				delete(c.groupsNotWatched, object.GroupVersion.Group)
				break
			}
		}
~~~

The Ingress is part of networking.k8s.io/v1beta1. Openshift 3.x supports networking.k8s.io/v1, but not v1beta1. As we compare the group name but not the version, `found` is set to `true`, which eventually leads to the panic.

LCaparelli:
Use a map indexed by the server group with another map as value indexed
by the versions of that group which are available in this cluster. This
allows for O(n) time complexity when checking if the desired objects are
available.

### DIFF
--- a/pkg/framework/controller_watcher.go
+++ b/pkg/framework/controller_watcher.go
@@ -95,7 +95,7 @@ func (c *controllerWatcher) Watch(watchedObjects ...WatchedObjects) (err error) 
 		if object.AddToScheme == nil {
 			desiredObjects = append(desiredObjects, object)
 		} else {
-			if _, found := serverGroupMap[object.GroupVersion.Group][object.GroupVersion.Version]; found {
+			if _, found := serverGroupMap[object.GroupVersion.String()]; found {
 				addToScheme = append(addToScheme, object.AddToScheme)
 				desiredObjects = append(desiredObjects, object)
 				delete(c.groupsNotWatched, object.GroupVersion.Group)
@@ -133,19 +133,18 @@ func (c *controllerWatcher) Watch(watchedObjects ...WatchedObjects) (err error) 
 	return
 }
 
-func (c *controllerWatcher) getServerGroupMap() (map[string]map[string]bool, error) {
+func (c *controllerWatcher) getServerGroupMap() (map[string]bool, error) {
 	serverGroups, err := c.discoverClient.ServerGroups()
 	if err != nil {
 		return nil, fmt.Errorf("couldn't fetch server groups list: %v", err)
 	}
 
-	serverGroupMap := map[string]map[string]bool{}
+	serverGroupMap := make(map[string]bool)
 	for _, serverGroup := range serverGroups.Groups {
-		versions := make(map[string]bool)
 		for _, version := range serverGroup.Versions {
-			versions[version.Version] = true
+			key := fmt.Sprintf("%s/%s", serverGroup.Name, version.Version)
+			serverGroupMap[key] = true
 		}
-		serverGroupMap[serverGroup.Name] = versions
 	}
 	return serverGroupMap, nil
 }
--- a/pkg/framework/controller_watcher.go
+++ b/pkg/framework/controller_watcher.go
@@ -18,12 +18,11 @@
 package framework
 
 import (
-	"strings"
-
+	"fmt"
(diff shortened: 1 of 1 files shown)


pkg/framework/controller_watcher.go
```diff
@@ -139,3 +132,20 @@ func (c *controllerWatcher) Watch(watchedObjects ...WatchedObjects) (err error)
...
+func (c *controllerWatcher) getServerGroupMap() (map[string]map[string]bool, error) {
+	serverGroups, err := c.discoverClient.ServerGroups()
+	if err != nil {
+		return nil, fmt.Errorf("couldn't fetch server groups list: %v", err)
+	}
+
+	serverGroupMap := map[string]map[string]bool{}
+	for _, serverGroup := range serverGroups.Groups {
+		versions := make(map[string]bool)
+		for _, version := range serverGroup.Versions {
```
ricardozanini, line 145: Instead of two maps, I think we could have only one with the `object.GroupVersion.String()` instead. wdyt? This way we'd decreased the complexity level as well. We could concatenate both of them with `fmt` like `version\group`. WDYT?

LCaparelli: I considered that when implementing it, but I was concerned with groups where the version wasn't necessarily at the end (`foo.bar/v1/foo`) which could make matching more difficult. We can definitely index the map with the help of`fmt`, good idea:

```
func (c *controllerWatcher) getServerGroupMap() (map[string]bool, error) {
	serverGroups, err := c.discoverClient.ServerGroups()
	if err != nil {
		return nil, fmt.Errorf("couldn't fetch server groups list: %v", err)
	}

	serverGroupMap := make(map[string]bool)
	for _, serverGroup := range serverGroups.Groups {
		for _, version := range serverGroup.Versions {
			key := fmt.Sprintf("%s/%s", serverGroup.Name, version.Version)
			serverGroupMap[key] = true
		}
	}
	return serverGroupMap, nil
}
```

The downside I see is that it would make access to the map more complex, as we'd need to build the key when accessing the map as well:

```
for _, object := range watchedObjects {
		// core resources
		if object.AddToScheme == nil {
			desiredObjects = append(desiredObjects, object)
		} else {
			key := fmt.Sprintf("%s/%s", object.GroupVersion.Group, object.GroupVersion.Version)
			if _, found := serverGroupMap[key]; found {
```

We could keep a const for format:

`const keyFormat = "%s/%s"`

What do you think? Indexing as `map[group/version]` or `map[group][version]`?

ricardozanini: When accessing the map, can't you just use `object.GroupVersion.String()` as key? That's the motivation behind creating the map like this. wdyt?

LCaparelli: Oh, yeah! I missed that, thanks :grin:

PR #52, opened 2020-05-07
Make Deployments use a ServiceAccount

### LINKED ISSUE #41: Pods can't be started on OCP 3.x clusters with default configuration
When attempting to deploy to an OCP 3.x cluster the following error pops up:

~~~
Error creating: pods "nexus3-695b67564f-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{200}: 200 is not an allowed group spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 200: must be in the ranges: [1000160000, 1000169999]]
~~~

As Nexus must run using this UID the cluster administrator needs to create a scc to work around this. It would be nice if we could supply this scc and have this documented.

LCaparelli:
Make Deployments use the ServiceAccount informed in
`nexus.spec.serviceAccountName`. If none is informed the default SA
created by the operator is used. 

Fix #41 by adding SCCs which can be used by the cluster admin and
documenting their use.


ricardozanini:
I have only one observation: since we are creating empty role and role binding, what’s the point creating them? We can just create the service account, bind it to the deployment and let the user customize the role with their needs. WDYT?

LCaparelli:
> I have only one observation: since we are creating empty role and role binding, what’s the point creating them? We can just create the service account, bind it to the deployment and let the user customize the role with their needs. WDYT?

Well, the user can still do that by creating another role that suits their needs and then binding to this account (as role policy rules are always additive I imagine there would be no conflict), but I do agree that the empty role seems a bit pointless.

I created it thinking we actually needed to bind the service account to a role, but I just ran a test creating only the service account and everything seems to be working fine. Gimme a few and I'll push a new commit :grin:

README.md
```diff
@@ -25,6 +25,41 @@ kubectl edit nexus
 ```
 
 If you're running on Kubernetes, edit the Nexus resource to add a [valid host for the Ingress](#network-on-kubernetes-114) to work.
+
+### Openshift 3.x
+
+If you're running the Operator on Openshift 3.x it's also necessary to configure a [Security Context Constraints](https://docs.openshift.com/container-platform/3.11/admin_guide/manage_scc.html) (SCC) resource.
+
+This is necessary because the Nexus image the operator uses requires its container to be ran as UID 200. The use of default SCC resources in Openshift 3.x results in a failure when starting the pods, as seen in [Issue #41](https://github.com/m88i/nexus-operator/issues/41).
```
ricardozanini, line 33:
```suggestion
This is necessary because the Nexus image requires its container to be ran as UID 200. The use of default SCC resources in Openshift 3.x results in a failure when starting the pods, as seen in [Issue #41](https://github.com/m88i/nexus-operator/issues/41).
```

README.md
```diff
@@ -25,6 +25,41 @@ kubectl edit nexus
...
+
+Valid SCC resources can be found at the `examples/` directory. You must associate the the SCC with the ServiceAccount in use. In the commands below it is assumed you'll be using the default ServiceAccount created when installing this Operator.
```
ricardozanini, line 35:
```suggestion
Valid SCC resources can be found at the `examples/` directory. You must associate the SCC with the `ServiceAccount` in use. In the commands below it is assumed you'll be using the default `ServiceAccount` created when deploying a new Nexus CR.
```

README.md
```diff
@@ -25,6 +25,41 @@ kubectl edit nexus
...
+
+> **Note**: you must choose one or the other, applying both will result in using the one applied last.
+
+Once the SCC has been created, run:
+
+```
+$ oc adm policy add-scc-to-user nexus-operator -z <ServiceAccountName>
+```
+
+This command will bind the SCC we just created with the ServiceAccount being used to create the Pods.
```
ricardozanini, line 57:
```suggestion
This command will bind the SCC we just created with the `ServiceAccount` being used to create the Pods.
```

README.md
```diff
@@ -121,6 +156,14 @@ drwxr-xr-x 19 root root  500 Apr 26 20:47 ..
 drwxr-xr-x  2  200  200 4096 Apr 26 15:42 pv0001
 ```
 
+## ServiceAccount
```
ricardozanini, line 159:
```suggestion
## Service Account
```

README.md
```diff
@@ -25,6 +25,41 @@ kubectl edit nexus
...
+
+Once the SCC has been created, run:
+
+```
+$ oc adm policy add-scc-to-user nexus-operator -z <ServiceAccountName>
+```
+
+This command will bind the SCC we just created with the ServiceAccount being used to create the Pods.
+
+If you're [using a custom ServiceAccount](##ServiceAccount), replace "`<ServiceAccountName>`" with the name of that account. If you're not using a custom ServiceAccount, the operator has created a default one which has the same name as your Nexus CR, replace "`<ServiceAccountName>`" with that.
```
ricardozanini, line 59:
```suggestion
If you're [using a custom ServiceAccount](##Service Account), replace "`<ServiceAccountName>`" with the name of that account. If you're not using a custom ServiceAccount, the operator has created a default one which has the same name as your Nexus CR, replace "`<ServiceAccountName>`" with that.
```

README.md
```diff
@@ -121,6 +156,14 @@ drwxr-xr-x 19 root root  500 Apr 26 20:47 ..
 drwxr-xr-x  2  200  200 4096 Apr 26 15:42 pv0001
 ```
 
+## ServiceAccount
+
+It is possible to use a custom [ServiceAccount](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/) to perform your Deployments with the Nexus Operator via:
```
ricardozanini, line 161:
```suggestion
It is possible to use a custom [`ServiceAccount`](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/) to perform your Deployments with the Nexus Operator via:
```

README.md
```diff
@@ -121,6 +156,14 @@ drwxr-xr-x 19 root root  500 Apr 26 20:47 ..
...
+
+  - `spec.serviceAccountName` (*string*): ServiceAccountName is the name of the ServiceAccount to use to run the Pods. If left blank, a default ServiceAccount is created with the same name as the Nexus CR.
```
ricardozanini, line 163:
```suggestion
  - `spec.serviceAccountName` (*string*): `ServiceAccountName` is the name of the `ServiceAccount` to run with the Pods. If left blank, a default `ServiceAccount` is created with the same name as the Nexus CR.
```

README.md
```diff
@@ -121,6 +156,14 @@ drwxr-xr-x 19 root root  500 Apr 26 20:47 ..
...
+
+**Important**: the Operator handles the creation of default resources necessary to run. If you choose to use a custom ServiceAccount be sure to also configure [Role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and [RoleBinding](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding) resources.
```
ricardozanini, line 165:
```suggestion
**Important**: the Operator handles the creation of default resources necessary to run. If you choose to use a custom `ServiceAccount` be sure to also configure [`Role`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and [`RoleBinding`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding) resources.
```

pkg/controller/nexus/resource/comparator.go
```diff
@@ -37,12 +39,18 @@ func GetComparator() compare.MapComparator {
...
+	roleBindingType := reflect.TypeOf(k8srbac.RoleBinding{})
+	svcAccntType := reflect.TypeOf(v1.ServiceAccount{})
 
 	resourceComparator.SetComparator(pvcType, resourceComparator.GetDefaultComparator())
 	resourceComparator.SetComparator(svcType, resourceComparator.GetComparator(svcType))
 	resourceComparator.SetComparator(deploymentType, resourceComparator.GetComparator(deploymentType))
 	resourceComparator.SetComparator(routeType, resourceComparator.GetComparator(routeType))
 	resourceComparator.SetComparator(ingressType, ingressEqual)
+	resourceComparator.SetComparator(roleType, rbac.GetComparator(roleType))
+	resourceComparator.SetComparator(roleBindingType, rbac.GetComparator(roleBindingType))
```
ricardozanini, line 52: These comparators can be removed

LCaparelli: Oops, nice catch hehe

PR #58, opened 2020-05-12
Add Table of Contents

ricardozanini:
See $Subject


ricardozanini:
Well, theoretically yes. Won't be easy because we need to replace the old one with the generated, and the tool does not do that: https://github.com/ekalinin/github-markdown-toc

One way is to add a markup like `<!-- begin toc -->`  and `<!-- end toc -->` to replace the contents by the generated one.

But I don't feel like doing this right now: #59 

hehehe

PR #63, opened 2020-05-15
update roles.yaml

nwalens:
As a fix for #41 the operator now creates service accounts based on parameters or Custom Resource names. For that behaviour to be correct, the roles have to include the required permissions.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -17,6 +17,7 @@ rules:
       - secrets
       - replicationcontrollers
       - podtemplates
+      - serviceaccounts
     verbs:
       - create
       - delete


ricardozanini:
I'll run some tests and update the CSV in your PR so we can launch 0.2.1 ASAP. Have you tested in which env?

ricardozanini:
I'll merge this one and send another with the other updates in the manifest. Thanks again!

nwalens:
sorry didn't see the reply during the weekend.  I had tested the role in a Openshift 4.3.8 cluster and it seemed to work.

Thanks a lot!!

Issue #64, opened 2020-05-15
Nexus Operator fails to create a ServiceAccount for the deployed CR


ricardozanini:
See: #63 

This is a regression from #41 fix.

In version 0.2.0 users will see the following message in the logs:
```
E0515 16:51:07.295411       1 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.4/tools/cache/reflector.go:105: Failed to list *v1.ServiceAccount: serviceaccounts is forbidden: User "system:serviceaccount:nexus:nexus-operator" cannot list resource "serviceaccounts" in API group "" in the namespace "nexus"
```

To workaround this issue the `nexus-operator` service account must have permissions on the object `serviceAccount`. We control this kind of permission in the `roles.yaml` file. See #41.

There are some other things to polish before to fix this bug. That's why we opened this issue.

PR #66, opened 2020-05-15
Running e2e with operator image and RBAC instead of local

ricardozanini:

This is an attempt to run the e2e tests with operator RBAC and a built image instead of local with the cluster admin user.

Having this would prevent errors like #64


ricardozanini:
Let's include in the 0.2.1 since there's no modification in the operator code itself and it's related to the 0.2.1 release in the first place.

PR #70, opened 2020-05-18
Closes #3: Add an option to not generate a random password for admin users 

### LINKED ISSUE #3: Add an option to set admin default password for new installations
Instead of relying on the image to set a random password and then have the user to `cat` the generated password in the container, would be better if we just set the default password on installations. The password can be changed later by the user manually.

ricardozanini:
See #3

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Table of Contents
       * [Persistence](#persistence)
          * [Minikube](#minikube)
       * [Service Account](#service-account)
-      * [Control random admin password generation](#control-random-admin-password-generation)
+      * [Control Random Admin Password Generation](#control-random-admin-password-generation)
       * [Red Hat Certified Images](#red-hat-certified-images)
 
 # Nexus Operator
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -16,6 +16,7 @@ Table of Contents
       * [Persistence](#persistence)
          * [Minikube](#minikube)
       * [Service Account](#service-account)
+      * [Control random admin password generation](#control-random-admin-password-generation)
```
LCaparelli, line 19:
```suggestion
      * [Control Random Admin Password Generation](#control-random-admin-password-generation)
```

We should capitalize this one in order to maintain consistency with the other headers.

README.md
```diff
@@ -182,6 +183,18 @@ It is possible to use a custom [`ServiceAccount`](https://kubernetes.io/docs/ref
 
 **Important**: the Operator handles the creation of default resources necessary to run. If you choose to use a custom ServiceAccount be sure to also configure [`Role`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and [`RoleBinding`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding) resources.
 
+## Control random admin password generation
```
LCaparelli, line 186:
```suggestion
## Control Random Admin Password Generation
```
We should capitalize this one in order to maintain consistency with the other headers.

PR #72, opened 2020-05-19
Integrating with GolangCI Lint

ricardozanini:
Enhancing our CI with [GolangCI Lint](https://golangci-lint.run/) integration. Also added a small contribution guide.


.github/ISSUE_TEMPLATE.md/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
```
LCaparelli, line 20: What do you think of also asking if the proposer if they would be available to test the feature in their environment? Something like:

```
**Would you be able to assist in testing this functionality if implemented?**
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
+# Contributing to Nexus Operator
+
+## Have found a bug or have a feature request?
+
+Please, open an issue for us. That will really help us to improve and also others that are using our project. There are templates ready for bug reporting and feature request to make things easier for you.
```
LCaparelli, line 5:
```suggestion
Please, open an issue for us. That will really help us improving the Operator and it would benefit other users such as yourself. There are templates ready for bug reporting and feature request to make things easier for you.
```
Wdyt?

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Have any questions?
+
+We're happy to answer! Either [open an issue](https://github.com/m88i/nexus-operator/issues) or send an email to our mailing list: [nexus-operator@googlegroups.com](mailto:nexus-operator@googlegroups.com).
+
+## Are willing to send a PR?
```
LCaparelli, line 11:
```suggestion
## Are you willing to send a PR?
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
...
+
+Before sending a PR, consider opening an issue first. So we can discuss your approach and the motivations behind it.
```
LCaparelli, line 13:
```suggestion
Before sending a PR, consider opening an issue first. This way we can discuss your approach and the motivations behind it while also taking into account other development efforts.
```

Wdyt?

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
...
+
+Regarding your local development environment:
+
+1. We use [golint-ci](https://golangci-lint.run/) to check the code. Consider [integrating it in your favorite IDE](https://golangci-lint.run/usage/integrations/) to avoid failing in the CI
+2. **Always** run `make test` before sending a PR to make sure the license headers and the manifests are updated (and of course the unit tests are passing)
+3. Consider adding a new [end-to-end](https://sdk.operatorframework.io/docs/golang/e2e-tests/) test case covering your scenario and make sure to run `make test-e2e` before sending the PR
```
LCaparelli, line 19:
```
4. Make sure to always keep your version of `Go` and the `operator-sdk` on par with the project. The current version information may be found at [our README pre-requisites section](https://github.com/m88i/nexus-operator#pre-requisites)
```

Wdyt? We still don't have that info on our pre-requisites section though. If you agree with this suggestion, would you mind putting that on the README as well?

go.mod
```diff
@@ -3,15 +3,14 @@ module github.com/m88i/nexus-operator
...
 	github.com/go-logr/logr v0.1.0
-	github.com/go-openapi/spec v0.19.7
+	github.com/go-openapi/spec v0.19.8
 	github.com/openshift/api v3.9.0+incompatible
 	github.com/operator-framework/operator-sdk v0.17.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.5.1
 	go.etcd.io/etcd v0.0.0-20191023171146-3cf2f69b5738
-	go.uber.org/zap v1.14.1
+	go.uber.org/zap v1.15.0
```
LCaparelli, line 12: Were these version bumps intentional? Are they required by GolangCI?

hack/go-lint.sh
```diff
@@ -16,14 +16,8 @@
 #     You should have received a copy of the GNU General Public License
 #     along with Nexus Operator.  If not, see <https://www.gnu.org/licenses/>.
 
+# it's expected to have GOPATH in your PATH
+# see this link to understand why we are using a different go.mod file: https://github.com/golang/go/issues/34506
+which golangci-lint >/dev/null || go get -modfile=go.tools.mod github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0
```
LCaparelli, line 21: This is pretty cool! I didn't know we could do that with `-modfile`.

Why are we specifying the version here though? I mean, we're already specifying the dependency and its version on `go.tools.mod`, so keeping it here as well would result in needing to update both `go.tools.mod` and this command whenever we bump versions.

It could also lead to this command overwriting `go.tools.mod` if we forget to change it here.

```suggestion
 which golangci-lint >/dev/null || go get -modfile=go.tools.mod github.com/golangci/golangci-lint/cmd/golangci-lint
```

Wdyt?

pkg/controller/nexus/resource/resources.go
```diff
@@ -37,7 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
LCaparelli, line 41: Isn't this the same as not renaming the import? Also, while we're at it, `v1`, `v1beta1` and so on are very cryptic imo. I think we should always use the package name and version when renaming, like `appsv1` above. Wdyt?

```suggestion
	corev1 "k8s.io/api/core/v1"
```

This would require changing its usages below as well, of course.

pkg/controller/nexus/resource/route.go
```diff
@@ -19,8 +19,9 @@ package resource
 
 import (
 	"fmt"
+
 	"github.com/m88i/nexus-operator/pkg/apis/apps/v1alpha1"
-	"github.com/openshift/api/route/v1"
+	v1 "github.com/openshift/api/route/v1"
```
LCaparelli, line 24: Same thing here. Let me know your thoughts!

```suggestion
	routev1 "github.com/openshift/api/route/v1"
```

pkg/controller/nexus/resource/route.go
```diff
@@ -33,7 +34,6 @@ const (
 type routeBuilder struct {
 	route *v1.Route
 	err   error
-	nexus *v1alpha1.Nexus
```
LCaparelli, line 36: Nice! This was one of the things I mentioned I wanted to change, completely unnecessary, but I missed it the first time hahaha :tada: :grin:

hack/go-lint.sh
ricardozanini: We must add the version, otherwise `go get` will take the last commit from that repo, which is not something that we want.

I agree that we should have a way to automate this version update, but won't cause any problems not updating it. The only thing is that we have to update the CI and the local script every time.

pkg/controller/nexus/resource/resources.go
ricardozanini: I agree, but that was changed by the lint/IDE itself. I will take this into account when refactoring/implementing new features to the code in the future. Let's leave that as is for now.

pkg/controller/nexus/resource/route.go
ricardozanini: Same comment as above.

pkg/controller/nexus/resource/route.go
ricardozanini: Lint remembered. hahaha

go.mod
ricardozanini: Was a suggestion from the lint tool. We were using a deprecated method and not handling an error propagation. Since I had to modify the code, I've decided to also update the library.

CONTRIBUTING.md
ricardozanini: It's worth mentioning here. I'll update the doc.

.github/ISSUE_TEMPLATE.md/feature_request.md
ricardozanini: Makes sense, I'll add it to the template. Thanks! :)

go.mod
LCaparelli: Ah, ok o/

hack/go-lint.sh
LCaparelli: Oh right, it downloads the latest release. I wasn't testing it correctly when I proposed the change. I suppose we need to keep it this way then.

ricardozanini:
As soon as Quay is back, we can run the checks again and merge this one.

PR #74, opened 2020-05-19
Fix Ingress comparison

### LINKED ISSUE #73: Ingress comparison is not working as expected
Ingress comparison is not working as expected due to 2 issues:

  1. the type registered to the comparator is `k8s.io/api/extensions/v1beta1.Ingress` instead of `k8s.io/api/networking/v1beta1.Ingress`
  2. the comparison function compares the deployed resource with itself, instead of comparing it with the required resource

LCaparelli:
Also removes unnecessary comparator function assignments for resources
which operator-utils knows how to handle.

Fixes #73 


ricardozanini:
Ready for rebase :)

PR #79, opened 2020-05-21
Rewrite resource life cycle logic

### LINKED ISSUE #55: Refactor resource management to simplify its logic
We could adopt the same pattern used in #52 to simplify the resource life cycle management. We could introduce an `ìnterface` along the lines of:

```go
type ResourceManager interface {
    GetRequiredResources() (map[reflect.Type][]resource.KubernetesResource, error)
    GetDeployedResources() (map[reflect.Type][]resource.KubernetesResource, error)
    GetComparator(t reflect.Type) func(deployed resource.KubernetesResource, requested resource.KubernetesResource)
}
```

Objects that implement this interface would be responsible for managing a domain of resources life cycle.

It could also simplify the logic contained in `pkg/controller/nexus/resource/resources.go` by storing a set of `ResourceManager` objects which could be iterated over. In pseudo-code, for example:

```
GetDeployedResources() {
    for manager in ResourceManagers {
        deployedResources += manager.GetDeployedResources()
    }
    return deployedResources
}
```

LCaparelli:
Rewrite resource life cycle logic by adding managers responsible for
creating, comparing and fetching deployed resources. The managers are
organized into 4 domains: persistence, networking, security and
deployment.

These managers implement an interface so that they can be easily
iterated over.

Fixes #55.


pkg/controller/nexus/resource/deployment/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const mgrNotInit = "the manager has not been initialized"
+
+var log = logger.GetLogger("deployment_manager")
+
+// Manager is responsible for creating deployment-related resources, fetching deployed ones and comparing them
+type Manager struct {
```
ricardozanini, line 39: Do we need to export this `struct`?
You may keep it private and create an `interface` instead. So the `NewManager` function would return the `manager` struct, that would implement the interface `Manager`. (or if `Manager` already implements the interface you created in another package, just keep it private since it's a structure that will only be used/created in here, and the builder `NewManager` is already handling it)

pkg/controller/nexus/resource/networking/manager.go
```diff
@@ -0,0 +1,208 @@
...
+const (
+	discFailureFormat    = "unable to determine if %s are available: %v" // resource type, error
+	resUnavailableFormat = "%s are not available in this cluster"        // resource type
+	mgrNotInit           = "the manager has not been initialized"
+)
+
+var log = logger.GetLogger("networking_manager")
+
+// Manager is responsible for creating networking resources, fetching deployed ones and comparing them
+type Manager struct {
```
ricardozanini, line 47: Same here regarding the exported `Manager` struct.

pkg/controller/nexus/resource/persistence/manager.go
```diff
@@ -0,0 +1,104 @@
...
+var log = logger.GetLogger("persistence_manager")
+
+// Manager is responsible for creating persistence resources, fetching deployed ones and comparing them
+type Manager struct {
+	nexus  *v1alpha1.Nexus
+	client client.Client
+}
+
+// NewManager creates a persistence resources manager
+func NewManager(nexus *v1alpha1.Nexus, client client.Client) *Manager {
```
ricardozanini, line 44: Same here.

pkg/controller/nexus/resource/resources.go
```diff
@@ -19,41 +19,55 @@ package resource
...
-	// CreateRequiredResources will create the requests resources as it's supposed to be
-	CreateRequiredResources(nexus *v1alpha1.Nexus) (resources map[reflect.Type][]resource.KubernetesResource, err error)
+	GetDeployedResources() (resources map[reflect.Type][]resource.KubernetesResource, err error)
+	// GetRequiredResources will create the requests resources as it's supposed to be
+	GetRequiredResources() (resources map[reflect.Type][]resource.KubernetesResource, err error)
+	// GetComparator returns the comparator based on the active managers
+	GetComparator() (compare.MapComparator, error)
+}
+
+type Manager interface {
```
ricardozanini, line 54: Oh here's the interface! :D

pkg/controller/nexus/resource/security/manager.go
```diff
@@ -0,0 +1,100 @@
...
+	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const mgrNotInit = "the manager has not been initialized"
+
+var log = logger.GetLogger("security_manager")
+
+// Manager is responsible for creating security resources, fetching deployed ones and comparing them
+type Manager struct {
```
ricardozanini, line 38: Same here

pkg/controller/nexus/resource/deployment/manager.go
LCaparelli: Oh, nicely spotted! I definitely agree, thanks!

PR #87, opened 2020-06-08
Move validation/defaults logic to Resource Managers (fix #82)

### LINKED ISSUE #82: Move all Nexus CR validations to the resource Managers
At this moment validations are spread out in the reconcile loop and in resource-generating functions. This makes the code harder to maintain as we need to worry about validation in more than one place. Additionally, projects importing this are not able to send a Nexus CR to the managers without making additional checks themselves.

Make the validation take place in a single place behind exposed API, improving code maintainability and usability.

Shift the validation responsibility to the Resource Managers. The reconcile loop would use them as would any users importing us as a library. The resource Managers would then be responsible for:

  - resource creation;
  - resource fetching;
  - resource comparison;
  - Nexus CR validation;

If the Manager finds an error it can't recover from when validating the Nexus CR it should return an error and die.

**Additional context**
N/A

Absolutely. :grin:

LCaparelli:
Most notable changes:

  - Move validation/default setting logic to the manager, centralizing
  it on a single exposed API
  - Make the `NewManager` functions receive a `v1alpha.Nexus` instead of
   a pointer to one. This is necessary because the newly introduced
   `setDefaults` methods destructively modify the nexus. We wouldn't
   want the API users having their original Nexus struct modified
   directly, so we pass a copy to the manager instead.
  - Wrap the `NewFakeDiscoveryClient` logic with a builder pattern so
  that the caller does not need to be aware unrelated abilities (e.g.,
  knowing if Ingresses are available on an Openshift cluster)

Fixes #82 


pkg/controller/nexus/resource/persistence/manager.go
```diff
@@ -41,11 +44,20 @@ type manager struct {
 }
 
 // NewManager creates a persistence resources manager
-func NewManager(nexus *v1alpha1.Nexus, client client.Client) *manager {
-	return &manager{
-		nexus:  nexus,
+func NewManager(nexus v1alpha1.Nexus, client client.Client) *manager {
```
ricardozanini, line 47: `manager` is implementing a public `interface`? I believe we should coordinate with #88

ricardozanini: Actually, `NewManager` must return `Manager` interface

LCaparelli:
Depends on #90

PR #88, opened 2020-06-09
Enable golint linter in golangci-lint

Kaitou786:
Enabled the golint linter to run with golangci-lint.
Signed-off-by: Tarun Khandelwal <tarkhand@redhat.com>

### DIFF
--- a/.github/workflows/nexus-operator-integration-checks.yml
+++ b/.github/workflows/nexus-operator-integration-checks.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v1.27
-          args: --timeout=2m
+          args: --enable=golint --timeout=2m
 
   unit_test:
     name: Unit Tests
--- a/pkg/test/client.go
+++ b/pkg/test/client.go
@@ -36,7 +36,7 @@ const (
 	monitoringGroupVersion = "monitoring.coreos.com/v1alpha1"
 )
 
-type fakeDiscBuilder struct {
+type FakeDiscBuilder struct {
 	*discfake.FakeDiscovery
 }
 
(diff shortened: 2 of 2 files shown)


Kaitou786:
@ricardozanini I just added the another package to call the functions instead of `resource.go` . I think it is preferred for a parent package to not call functions from its child package, So I created a sibling package and did these things and only left the parent package with the interfaces. But yes it can be in a same package

ricardozanini:
@Kaitou786 understood! Let's wait for @LCaparelli, I believe he will fix this in his PR #87. Then we can rebase and merge this one. :)

ricardozanini:
@LCaparelli from the client perspective (`NexusController`) how it will be? `deployment.NewManager()`? Or `resources.New()` (all managers)? 

I'm not sure if I like the name `interfaces`, we could use `infra` instead. Plus, it's not a good practice having packages in plural, so I'd say to also rename `resources` to `resource`.

LCaparelli:
> @LCaparelli from the client perspective (`NexusController`) how it will be? `deployment.NewManager()`? Or `resources.New()` (all managers)?

That would be left unchanged, the clients would still instantiate the `NexusResourceManager` (or `Supervisor`), which would in turn instantiate the Managers via their respective package, such as `deployment.NewManager`.

Someone using the API could also instantiate the Managers separately, without the need for a Supervisor, by running `deployment.NewManager` or `networking.NewManager` (or whichever manager) directly.

Anyone using either the Managers or the Supervisor would still need to import `interface` though, in order to manipulate these structs. 

> I'm not sure if I like the name `interfaces`, we could use `infra` instead.

I agree that `interface` could also carry more meaning. I'm still not sure if there isn't some other name we could use that is more significant, but I'm definitely ok with `infra` as well.

> Plus, it's not a good practice having packages in plural, so I'd say to also rename `resources` to `resource`.

Agreed :+1:

Kaitou786:
@ricardozanini @LCaparelli do we want to make these changes as a part of this PR, or as a part of PR #87 ?

ricardozanini:
> @ricardozanini @LCaparelli do we want to make these changes as a part of this PR, or as a part of PR #87 ?

@Kaitou786 I think @LCaparelli forgot to mention, he's working on a new PR to fix those, so we can later rebase this one and merge it. Probably tomorrow on your TZ. :)

We will let you know during your sleep :p

LCaparelli:
@Kaitou786 I think this would be a better fit for a new PR altogether, I'll send it today. Once we merge that we can rebase it for this and #87. This way each of our PRs only solve one issue :-)

Kaitou786:
Ah okay, thank you guys =D

LCaparelli:
Depends on #90

ricardozanini:
@Kaitou786 just rebase and let's be happy. :D

Kaitou786:
@LCaparelli exposed the `fakeDiscBuilder` and it fixed all, thanks :D

PR #97, opened 2020-06-19
Allow the user to select the Pull Policy

### LINKED ISSUE #83: Allow the user to select the Pull Policy
At the moment the Pull Policy is hardcoded as "Always", meaning the user has no choice whether or not to pull an image.

It would be nice if the user could select the policy as they see fit.

There could be a new field in the Nexus CRD to control this.

**Additional context**
N/A

Yes :grin:

LCaparelli:
Fix #83 


LCaparelli:
@Kaitou786 in case you want to have a look too :-)

Kaitou786:
@LCaparelli thanks, these changes look good to me =D, just one thing do you think there should be a section in `README` about this `imagePullPolicy` so user can know?

LCaparelli:
Yup that's true. I'll document it in a bit, thanks!

README.md
```diff
@@ -209,6 +210,18 @@ If you have access to [Red Hat Catalog](https://access.redhat.com/containers/#/r
...
 
+## Image Pull Policy
+
+You can control the pods Image Pull Policy using the `spec.imagePullPolicy` field. It accepts either of the following values:
+
+  - `Always`
+  - `IfNotPresent`
+  - `Never` 
+
+If this field is set to an invalid value this configuration will be omitted, deferring the behavior to Kubernetes default, which is `Always` if the image's tag is "latest" and `IfNotPresent` otherwise.
```
ricardozanini, line 221: Do we have a reference for this statement (like a Kube doc)? I remember @Kaitou786 brought this up, maybe he has it? :)

deploy/olm-catalog/nexus-operator/0.3.0/nexus-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -75,6 +75,13 @@ spec:
         path: image
         x-descriptors:
         - urn:alm:descriptor:io.kubernetes:image
+      - description: 'The image pull policy for the Nexus image. If left blank behavior
+          will be determined by the image tag (`Always` if "latest" and `IfNotPresent`
+          otherwise). Possible values: `Always`, `IfNotPresent` or `Never`.'
+        displayName: Image Pull Policy
+        path: imagePullPolicy
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:imagePullPolicy
```
ricardozanini, line 84: Oh, I didn't know we have this `x-descriptor` type. 
EDIT: it seems the operator-sdk is now smarter than me =D

README.md
LCaparelli: @ricardozanini From K8s all I found was [this](https://kubernetes.io/docs/concepts/containers/images/#updating-images), which kind of breezes over the matter, but I think is clear enough.

On the other hand, [Openshift's doc](https://docs.openshift.com/container-platform/3.6/dev_guide/managing_images.html#image-pull-policy) gives it more attention, with its own section (though the conclusions we can draw from both are ultimately the same, of course).

I suppose we can link the K8s one, unless @Kaitou786 actually saw this somewhere else.

ricardozanini: +1 to add the Kuberentes ref. :)

Kaitou786: The kubernetes ref seems better, origanlly I read it in comments in the `Container Struct` where this is defined

ricardozanini: @LCaparelli as soon as you update the docs, we can merge this.

Kaitou786:
Just a question, if the user defines any wrong value in `imagePullPolicy` i.e. not in [`Always`, `IfNotPresent` `Never`]. Our Manager will set the value to empty string, I think that'll set the image pull policy to Always regardless of any tag of the image. Do we want this behavior?

LCaparelli:
@Kaitou786 I just tested this using a custom image with a tag that's not "latest" or that isn't omitted and the Pull Policy becomes `IfNotPresent`:

```
╰─ kubectl -n issue83 get nexus -o yaml | grep ' image'      
    image: my-custom-image:mytag
    imagePullPolicy: invalid
╰─ kubectl -n issue83 get deployment -o yaml | grep ' image' 
          image: my-custom-image:mytag
          imagePullPolicy: IfNotPresent
```

I think maybe you were under that impression because the default images have the "latest" tag (causing it to set it to `Always`)?

The reason why we're setting that to an empty string is because that's Go's zero value for strings, so it's as if it had never been defined and it receives the value it would if we hadn't assigned anything to it in the first place. If that was a pointer to a string, on the other hand, we would have set it `nil` instead.

Kaitou786:
@LCaparelli that's nice then, I just got confused from the last line in https://kubernetes.io/docs/concepts/containers/images/#updating-images 
`When imagePullPolicy is defined without a specific value, it is also set to Always.`

LCaparelli:
Oh that's well spotted though @Kaitou786. I think it only applies to YAML or JSON fed into the API, as in this case you *can* omit the key altogether, which we can't in Go. :-)

LCaparelli:
Alright @ricardozanini @Kaitou786 the doc ref is in :-D :tada:

PR #98, opened 2020-06-24
Add Nexus3 health check using its own API endpoint

### LINKED ISSUE #80: Add Nexus3 health check using its own API endpoint
Today the health check just makes a HTTP request to the default port, instead the Nexus Server offers a standard health check endpoint on `/service/rest/v1/status`.

The Nexus deployment probes to check the default health endpoint instead of the default port only.

None

**Additional context**
See the documentation: https://help.sonatype.com/repomanager3/rest-and-integration-api/status-api#StatusAPI-Status

Yes

Kaitou786:
Changed Readiness and Liveliness Probe to check its own status API
Closes #80

### DIFF
--- a/pkg/controller/nexus/resource/deployment/deployment.go
+++ b/pkg/controller/nexus/resource/deployment/deployment.go
@@ -115,7 +115,7 @@ func addProbes(nexus *v1alpha1.Nexus, deployment *appsv1.Deployment) {
 	livenessProbe := &corev1.Probe{
 		Handler: corev1.Handler{
 			HTTPGet: &corev1.HTTPGetAction{
-				Path: "/",
+				Path: "/service/rest/v1/status",
 				Port: intstr.IntOrString{
 					IntVal: NexusServicePort,
 				},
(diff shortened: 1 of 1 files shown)


ricardozanini:
> IMO, we can keep this as it is (report success if reads are available) since the most common operation in a Nexus repository is reading from it.

Yes, I thought about that when I was reporting the issue and reading Nexus docs, I pondered that having the Nexus in `read` state would be enough for most cases, so won't be nice to have Kubernetes to redeploy the pod in such situations, but to continue offer the libraries a build process might need. 

What can be done is to read this state from the Nexus repo and report it as an Event in Kubernetes monitoring console. That would be something pretty cool to do, but we need [Aicura](https://github.com/m88i/aicura/) to be finished first (to call the internal Nexus API).

LCaparelli:
I see, I think that would be cool too! @Kaitou786 I'm approving the PR as it is, but let us know if you think that we should probe the `/service/rest/v1/status/writable` endpoint instead :grin:

Kaitou786:
I think we can leave it as it is. Also having that event on k8s dashboard would be pretty cool!

PR #107, opened 2020-07-02
Allow the fake client to mock server API error responses

### LINKED ISSUE #105: Allow the fake client to mock server API error responses
Today we have the following implementation for the fake client used in testing, taken from `pkg/test/client.go`:

```go
// NewFakeClient will create a new fake client with all needed schemas
func NewFakeClient(initObjs ...runtime.Object) client.Client {
	return fake.NewFakeClientWithScheme(GetSchema(), initObjs...)
}

// GetSchema gets the needed schema for fake tests
func GetSchema() *runtime.Scheme {
	s := scheme.Scheme
	s.AddKnownTypes(v1alpha1.SchemeGroupVersion, &v1alpha1.Nexus{})
	s.AddKnownTypes(routev1.GroupVersion, &routev1.Route{}, &routev1.RouteList{})
	return s
}
```

While this is nice and really useful, it doesn't allow us to mock specific responses to certain events/actions. That feature would come in pretty handy when testing the resource managers' ability to handle/report errors when the server responds, for example, with a 500 status code.

Digging up some docs I found that `k8s.io/client-go/testing` (which is already a dependency for the project) has a nice fake client implementation that allow us to insert "interceptors" (called `Reactor` here) that allow us to define how the fake client should respond to certain actions.

Check out its [godoc](https://godoc.org/k8s.io/client-go/testing) for all the good stuff and [this issue](https://github.com/kubernetes/client-go/issues/500) for an example of working usage.

Also noteworthy is that our fake discovery client uses the `k8s.io/client-go/discovery/fake` package, which is built on top of the same `Fake` [implementation](https://github.com/kubernetes/client-go/blob/5b8ea8e61cb840f5c6f6baacd67ae93fe30a3027/testing/fake.go).

LCaparelli:
Encapsulate the fake API and discovery clients in a single struct
capable of mocking error responses that implements both relevant
interfaces ([client.Client](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L104) and [discovery.DiscoveryInterface](https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L55)).

Fix #105 

### DIFF
--- a/pkg/test/client.go
+++ b/pkg/test/client.go
@@ -18,67 +18,270 @@
 package test
 
 import (
+	"context"
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
 	"github.com/m88i/nexus-operator/pkg/apis/apps/v1alpha1"
 	routev1 "github.com/openshift/api/route/v1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/discovery"
 	discfake "k8s.io/client-go/discovery/fake"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	clienttesting "k8s.io/client-go/testing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 const (
-	openshiftGroupVersion  = "openshift.io/v1"
-	monitoringGroupVersion = "monitoring.coreos.com/v1alpha1"
+	openshiftGroupVersion = "openshift.io/v1"
 )
 
-type FakeDiscBuilder struct {
-	*discfake.FakeDiscovery
+// FakeClientBuilder allows building a FakeClient according to
+// the desired cluster capabilities
+type FakeClientBuilder struct {
+	initObjs  []runtime.Object
+	scheme    *runtime.Scheme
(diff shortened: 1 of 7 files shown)


Kaitou786:
After this PR, as our coverage is increasing, do you think we should increase our coverage target from 54 -> 58 as a part of this one only?

ricardozanini:
> After this PR, as our coverage is increasing, do you think we should increase our coverage target from 54 -> 58 as a part of this one only?

Totally agree. We should keep pushing forward as long as the coverage increases. :)

pkg/test/client.go
```diff
@@ -18,67 +18,270 @@
...
 }
 
-// NewFakeClient will create a new fake client with all needed schemas
-func NewFakeClient(initObjs ...runtime.Object) client.Client {
-	return fake.NewFakeClientWithScheme(GetSchema(), initObjs...)
+// NewFakeClientBuilder will create a new fake client that is aware of minimal resource types
+// and stores initObjs for initialization later
+func NewFakeClientBuilder(initObjs ...runtime.Object) *FakeClientBuilder {
+	s := scheme.Scheme
+	s.AddKnownTypes(v1alpha1.SchemeGroupVersion, &v1alpha1.Nexus{}, &v1alpha1.NexusList{})
```
ricardozanini, line 54: I’m on the phone. But take a look at the Kogito Operator, I’ve changed the way we add schemas there without having to add objects manually like this.

Kaitou786:
@LCaparelli I think you forgot to change the target for codecov. It needs to be updated on the `codecov.yml`. Thanks =D

LCaparelli:
Ugh, yes I did! Thanks for pointing it out @Kaitou786, I got carried away studying the scheme stuff and forgot about it. New commit on the way

PR #108, opened 2020-07-03
Remove unnecessary hack

LCaparelli:


LCaparelli:
For whatever reason I had thought this was necessary. It isn't, as interfaces don't enforce a particular type of receiver, meaning I could have just changed these methods to pointer receivers instead of introducing this hack.

My bad. :-/

Issue #101, opened 2020-06-24
Add unit tests to Resource Managers


LCaparelli:
There are no unit tests for the Resource Managers, it would be nice to have some.

PR #110, opened 2020-07-06
Refactor resource managers to return their struct, not the interface

### LINKED ISSUE #109: Refactor resource managers to return their struct, not the interface
When creating a new resource manager today we return its interface:

```go
func NewManager(nexus v1alpha1.Nexus, client client.Client) infra.Manager {
```

While this is fine at the moment, it goes against the ["Accept Interfaces Return Struct"](https://mycodesmells.com/post/accept-interfaces-return-struct-in-go) rule of thumb. Returning an interface brings no additional benefit, but it can reduce the code's flexibility by removing all of the object's behavior that's not explicitly defined in the interface.

For example, consider the following interface:

```go
type Person interface {
    Name() string
    Age() int
}
```

One implementation of that interface could be:

```go
type Gamer struct {
	name  string
	age   int
	games []Game
}

func (g Gamer) Name() string {
	return g.name
}

func (g Gamer) Age() int {
	return g.age
}

func (g Gamer) Games() []Game {
	return g.games
}
```
If the function that creates a new gamer returned `Person`, we could never access this additional behavior defined in `Games()`. If it returned `*Gamer`, we'd still be able to use functions which receive a `Person` as parameter, as `Gamer` implements that interface AND we'd still be able to make use of this type's specific behavior.

LCaparelli:
Returning an interface can reduce the code's flexibility by removing all
of the object's behavior that's not explicitly defined in the interface.
By returning a struct instead, we leave it up to the caller to decide
how the return should be used.

Most notable changes:
  - Return the actual resource `Manager` structs instead of their interface
  - Expand on the networking manager's API to provide convenient means for
  users to determine if ingresses and routes are available.
  - Move resource `Supervisor` and `Manager` interfaces back to the 
  `resource` package as there is no longer a dependency cycle caused by 
  their presence there

Fix #109

### DIFF
--- a/pkg/controller/nexus/nexus_controller.go
+++ b/pkg/controller/nexus/nexus_controller.go
@@ -20,7 +20,6 @@ package nexus
 import (
 	"context"
 	"fmt"
-	"github.com/m88i/nexus-operator/pkg/controller/nexus/resource/infra"
 	"reflect"
 
 	"github.com/m88i/nexus-operator/pkg/cluster/kubernetes"
@@ -115,7 +114,7 @@ type ReconcileNexus struct {
 	client             client.Client
 	scheme             *runtime.Scheme
 	discoveryClient    discovery.DiscoveryInterface
-	resourceSupervisor infra.Supervisor
+	resourceSupervisor resource.Supervisor
 }
 
 // Reconcile reads that state of the cluster for a Nexus object and makes changes based on the state read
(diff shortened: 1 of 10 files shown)


LCaparelli:
>  The diff coverage is 38.29%.

This is lower than expected as this PR introduces nil checks to exposed methods when necessary. I'm currently working on unit tests for these methods on #101 and adding these on this PR will make it all very confusing. If that's ok with you guys I should be able to fix all of this and get in a goood bump on overall coverage soon when I push another PR for the tests. :-)

PR #114, opened 2020-07-08
Add unit tests to Resource Managers

### LINKED ISSUE #101: Add unit tests to Resource Managers

LCaparelli:
Depends on #112 
Fix #101

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,6 +6,6 @@ coverage:
   status:
     project:
       operator:
-        target: 58
+        target: 83
         flags:
           - operator
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,6 +6,7 @@ coverage:
   status:
     project:
       operator:
-        target: 83
+        target: auto
+        threshold: 5%
         flags:
           - operator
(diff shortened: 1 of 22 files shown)


codecov.yml
```diff
@@ -6,6 +6,6 @@ coverage:
   status:
     project:
       operator:
-        target: 58
+        target: 83
```
ricardozanini, line 9: We might keep on 80 only. :thinking:

LCaparelli: I was thinking of setting it to `auto` instead of keeping an actual value from now on, maybe with a 2% or 4% threshold. This way the target will be the one on the latest commit. Explained [here](https://docs.codecov.io/docs/commit-status).

Say coverage is currently at 83% and we're using a 4% threshold, `auto` will set the target to that and the new PR will pass the check as long as its coverage is > 79%

Wdyt?

Kaitou786: +1 for using this to auto and maybe setting the threshold to 5%

ricardozanini: +1 for auto and 5%. I didn't know that we could have this possibility. Go ahead!

PR #113, opened 2020-07-07
Create Make target for pre-PR checks and push automation

LCaparelli:
Signed-off-by: Lucas Caparelli <lucas.caparelli112@gmail.com>

PR #117, opened 2020-07-09
Let incorrect usage of the exposed API panic

### LINKED ISSUE #116: Let incorrect usage of the exposed API panic
The methods exposed by the resource managers have checks to make sure the struct has been correctly initialized and return an `error` if not. For example, in `pkg/controller/nexus/resource/deployment/manager.go`:

```go
// GetDeployedResources returns the deployment-related resources deployed on the cluster
func (m *Manager) GetDeployedResources() ([]resource.KubernetesResource, error) {
	if m.nexus == nil || m.client == nil {
		return nil, fmt.Errorf(mgrNotInit)
	}
```

These don't go in line with the error-handling philosophy in Go. The `error` type exists to report errors which are expected, possible outcomes of a correct usage of the function/method.

From "The Go Programming Language" (Donovan, Kernighan - 2015):

>Some functions always succeed at their task. For example, `strings.Contains` and `strconv.FormatBool` have well-defined results for all possible argument values and cannot fail — barring catastrophic and unpredictable scenarios like running out of memory, where the symptom is far from the cause and from which there’s little hope of recovery.

>Other functions always succeed so long as their preconditions are met. For example, the `time.Date` function always constructs a `time.Time` from its components — year, month, and so on — unless the last argument (the timezone) is nil, in which case it panics. This panic is a sure sign of a bug in the calling code and should never happen in a well-written program.

In the end, we shouldn't return an error here just to avoid a panic for nil dereference, we should embrace it as a clear way of letting the caller know that this is a consequence of incorrect usage and it's most definitely not an expected outcome from this method.

LCaparelli:
We shouldn't return an error here just to avoid a panic for nil
dereference, we should embrace it as a clear way of letting the caller
know that this is a consequence of incorrect usage and it's most
definitely not an expected outcome from these methods.

Fix #116

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,5 +8,6 @@ coverage:
       operator:
         target: auto
         threshold: 5%
+        base: auto
         flags:
           - operator
--- a/pkg/controller/nexus/resource/deployment/manager.go
+++ b/pkg/controller/nexus/resource/deployment/manager.go
@@ -36,7 +36,6 @@ import (
 const (
 	NexusCommunityLatestImage       = "docker.io/sonatype/nexus3:latest"
 	NexusCertifiedLatestImage       = "registry.connect.redhat.com/sonatype/nexus-repository-manager"
-	mgrNotInit                      = "the manager has not been initialized"
 	probeDefaultInitialDelaySeconds = int32(240)
 	probeDefaultTimeoutSeconds      = int32(15)
 	probeDefaultPeriodSeconds       = int32(10)
@@ -68,6 +67,7 @@ var (
 )
 
 // Manager is responsible for creating deployment-related resources, fetching deployed ones and comparing them
+// Use with zero values will result in a panic. Use the NewManager function to get a properly initialized manager
 type Manager struct {
 	nexus  *v1alpha1.Nexus
 	client client.Client
@@ -158,10 +158,6 @@ func (m *Manager) setProbeDefaults() {
 
 // GetRequiredResources returns the resources initialized by the manager
 func (m *Manager) GetRequiredResources() ([]resource.KubernetesResource, error) {
-	if m.nexus == nil || m.client == nil {
-		return nil, fmt.Errorf(mgrNotInit)
-	}
-
 	log.Debugf("Creating Deployment (%s)", m.nexus.Name)
 	deployment := newDeployment(m.nexus)
 	log.Debugf("Creating Service (%s)", m.nexus.Name)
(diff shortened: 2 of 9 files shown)


LCaparelli:
Hmmm, something is fishy. Why is the coverage target 62%? It was already 80 something on previous commits, no?

LCaparelli:
Screenshot from 2020-07-09 10-41-09

The "pseudo" commit it mentions is [77bf38c](https://github.com/m88i/nexus-operator/commit/77bf38c82458f4fdce38083b9ca7dc41944cfa8b) while the "base" is [1e7333c](https://github.com/m88i/nexus-operator/commit/1e7333c548ebfad54323fc8c5f6c6fc6a8f4ff8a), which is the README update from yesterday.
 
Screenshot from 2020-07-09 10-45-52
It seems like it skipped [aeec0e0](https://github.com/m88i/nexus-operator/commit/aeec0e08c6f258a7ffa15fa50968b05e7e4ada0d) for some reason. Maybe it'll work on the next PR?

Kaitou786:
> Screenshot from 2020-07-09 10-41-09
> 
> The "pseudo" commit it mentions is [77bf38c](https://github.com/m88i/nexus-operator/commit/77bf38c82458f4fdce38083b9ca7dc41944cfa8b) while the "base" is [1e7333c](https://github.com/m88i/nexus-operator/commit/1e7333c548ebfad54323fc8c5f6c6fc6a8f4ff8a), which is the README update from yesterday.
>  
> Screenshot from 2020-07-09 10-45-52
> It seems like it skipped [aeec0e0](https://github.com/m88i/nexus-operator/commit/aeec0e08c6f258a7ffa15fa50968b05e7e4ada0d) for some reason. Maybe it'll work on the next PR?

This is strange behaviour though, it should've taken the latest commit. I wonder why it got stuck at a different commit.

Kaitou786:
Do you think the `base` parameter should also be added in the codecov.yml file which can cause this issue? [Check here](https://docs.codecov.io/docs/commit-status)

LCaparelli:
> Do you think the `base` parameter should also be added in the codecov.yml file which can cause this issue? [Check here](https://docs.codecov.io/docs/commit-status)

I think it's worth a shot, nice catch. I'll send a new commit shortly.

LCaparelli:
@Kaitou786, @ricardozanini found [this issue](https://github.com/codecov/codecov-action/issues/43) which seems to be related

Kaitou786:
> @Kaitou786, @ricardozanini found [this issue](https://github.com/codecov/codecov-action/issues/43) which seems to be related

I'm not sure the reports seems to be uploaded on the codecov for those 2 PRs in between  see: [114](https://github.com/m88i/nexus-operator/pull/114/checks?check_run_id=850454036#step:9:46) [113](https://github.com/m88i/nexus-operator/pull/113/checks#step:9:46). Also, can you try specifying `parent` in the base field?

Kaitou786:
Maybe after defining the base, it'll work from next PR. In any case after this is merged with the base, I'll send another PR tomorrow morning and see if it is working fine :)

PR #119, opened 2020-07-10
change codecov to use action

Kaitou786:
And validate the yaml file

### DIFF
--- a/.github/workflows/nexus-operator-integration-checks.yml
+++ b/.github/workflows/nexus-operator-integration-checks.yml
@@ -62,9 +62,15 @@ jobs:
 
       - name: Test Packages
         run: go test ./pkg/... -count=1 -coverprofile=coverage.txt -covermode=atomic
-
+      - name: Validate codcov yaml file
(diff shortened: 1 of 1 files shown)


.github/workflows/nexus-operator-integration-checks.yml
```diff
@@ -62,9 +62,15 @@ jobs:
 
       - name: Test Packages
         run: go test ./pkg/... -count=1 -coverprofile=coverage.txt -covermode=atomic
-
+      - name: Validate codcov yaml file
+        run: cat codecov.yml | curl --data-binary @- https://codecov.io/validate
```
LCaparelli, line 66: I think we can simplify this to spawn a single process. We might also want to add verbosity so we can check what happened in case something goes wrong with the connection itself:

```suggestion
        run: curl -vvv --data-binary @codecov.yml https://codecov.io/validate
```

Kaitou786: Ah yes, this should work. I just copied the command from [here](https://codecov.io/gh/Kaitou786/nexus-operator/settings/yaml)

PR #121, opened 2020-07-13
Improve hack scripts

Kaitou786:


hack/ci/install-kind.sh
```diff
@@ -17,19 +17,21 @@
...
 
-if [ ! -f ./bin/kind ]; then
-    echo "Installing KIND"
-    curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-amd64
-    chmod +x ${PWD}/kind &&
-        mkdir -p ./bin &&
-        mv ${PWD}/kind ./bin/kind
+if [[ $(which kind) ]]; then
+  echo "---> kind is already installed. Please make sure it is the required ${KIND_VERSION} version before proceeding"
 else
```
LCaparelli, line 30: `which` should be avoided when possible. It spaws a different process and implementation may vary from one distribution to another, so it may not even set a return value which is used in the comparation here. In this case we can stick to the `command` built-in:
```suggestion
if ! command -v kind &> /dev/null; then
  echo "---> kind is already installed. Please make sure it is the required ${KIND_VERSION} version before proceeding"
else
```
More details [here](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script).

hack/ci/install-operator-sdk.sh
```diff
@@ -16,18 +16,23 @@
...
-if [ ! -f ./bin/operator-sdk ]; then
-    echo "Installing Operator SDK version ${OPERATOR_SDK_VERSION}"
-    curl -LO https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu
+GOPATH=$(go env GOPATH)
 
-    chmod +x ${PWD}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu &&
-        mkdir -p ./bin &&
-        cp ${PWD}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu ./bin/operator-sdk &&
-        rm ${PWD}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu
+if [[ $(which operator-sdk) ]]; then
```
LCaparelli, line 29: Same observation regarding `which`

hack/ci/start-kind.sh
```diff
@@ -18,24 +18,24 @@
 
 set -e
 
-if [ -z "${KUBE_VERSION}" ]; then
-    KIND_KUBE_VERSION=""
-else
-    KIND_KUBE_VERSION="--image kindest/node:${KUBE_VERSION}"
-fi
+#Make sure docker is installed before proceeding
+which docker > /dev/null || (echo "docker is not installed. Please install it before proceeding exiting...." && exit 1)
```
LCaparelli, line 22: Not trying to be obnoxious, but marking it so none are left behind :-)

You should still be able to keep a similar syntax:

```suggestion
command -v docker &> /dev/null || (echo "docker is not installed. Please install it before proceeding exiting...." && exit 1)
```

hack/ci/start-kind.sh
```diff
@@ -18,24 +18,24 @@
...
 
-if [ ! -z ${VERBOSE+x} ] && [ "${VERBOSE}" == "1" ]; then
-    KIND_VERBOSITY="--verbosity 1"
-else
-    KIND_VERBOSITY="--verbosity 0"
-fi
+#make sure kind is installed before proceeding
+which kind > /dev/null || (echo "kind is not installed. Please install it before proceeding exiting...." && exit 1)
```
LCaparelli, line 25: Here :-)

hack/go-vet.sh
```diff
@@ -16,18 +16,19 @@
...
-    operator-sdk generate crds
+operator-sdk generate k8s
+operator-sdk generate crds
 
-    # get the openapi binary
-    which ./bin/openapi-gen >/dev/null || go build -o ./bin/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen
-    echo "Generating openapi files"
-    ./bin/openapi-gen --logtostderr=true -o "" -i ./pkg/apis/apps/v1alpha1 -O zz_generated.openapi -p ./pkg/apis/apps/v1alpha1 -h ./hack/boilerplate.go.txt -r "-"
+# get the openapi binary
+which openapi-gen >/dev/null || go build -o $GOPATH/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen
```
LCaparelli, line 28: And finally here :-)

hack/ci/install-kind.sh
Kaitou786: Updated to use command, didn't use & at the end, it is working as expected without it

PR #123, opened 2020-07-14
Comment only when change in codecov report

Kaitou786:


Kaitou786:
According the documentation the comment should not be there, even if commented before it should be deleted as coverage is same [See](https://docs.codecov.io/docs/pull-request-comments#requiring-changes)

ricardozanini:
Well, let's merge either way. I don't think this option would cause a harm.

Kaitou786:
@ricardozanini the olm starts failing. I think could be related to the testing image which was recently updated: https://quay.io/repository/operator-framework/operator-testing?tag=latest&tab=tags. We are having this same error on kogito-cloud-operator

ricardozanini:
> @ricardozanini the olm starts failing. I think could be related to the testing image which was recently updated: https://quay.io/repository/operator-framework/operator-testing?tag=latest&tab=tags. We are having this same error on kogito-cloud-operator

I noticed, I opened #124 
Will fix it in a couple hours.

LCaparelli:
Shall we merge this @ricardozanini?

Kaitou786:
@ricardozanini do you think we can have similar approach we are using on the kogito-cloud-operator?

Kaitou786:
> > @ricardozanini do you think we can have similar approach we are using on the kogito-cloud-operator?
> 
> I think that we could skip running tests on docs changes, but I believe we already have that set, right?

Currenly we've set the following config:
```
on:
  pull_request:
  push:
    paths-ignore:
      - "docs/**"
      - "README.md"
      - "CONTRIBUTING.md"
      - "examples/**"
      - "LICENSE"
      - "LICENSE_NOTICE"
      - "Makefile"
      - ".github/ISSUE_TEMPLATE/**"
    branches:
      - master
```
Due to which for every PR, which is created we run the tests no matter where the changes. For push, we've ignored those changes. I noticed recently whenever we merge the PR on the master it runs our test again as we have that in our workflow. I think we should remove the push section as the tests are running in redundantly after the merge.  And add the paths-ignore in the pull section. 
^^ @ricardozanini @LCaparelli wdyt?

LCaparelli:
Sounds good to me @Kaitou786

ricardozanini:
@Kaitou786 are you going to send a PR to make these changes?

Kaitou786:
> @Kaitou786 are you going to send a PR to make these changes?

Added the commit as the part of this one only :-)

PR #133, opened 2020-07-29
Implement seamless updates for the community image

### LINKED ISSUE #132: Implement seamless upgrade for the community image
In order to provide seamless upgrade capabilities (see #62) we must provide it for the community Nexus use case. This should support both Kubernetes and Openshift, though without using ImageStreams yet.

When using the default community image, the Operator is able to automatically:
  - determine if a newer micro version of the image is available;
  - update to the newer version;
  - rollback the update if it fails;
  - provide humans with information about attempts (regardless if they are successful or not);

This will probably require a some changes to the Nexus CRD. It should give the user the ability to:
  - choose a Nexus minor version [1] (this should **not** have any effect if Red Hat certified images are in use for now);
  - disable automatic updates;

Automatic updates should not be available with custom images as we're unable to determine if they follow [semantic versioning](https://semver.org/). 

[1]: I was unable to find information on whether or not the community image follows semantic versioning. Keeping on the same minor is a conservative approach to avoid compatibility issues. If semantic versioning is correctly followed, the user wouldn't need specify any versions as the Operator supports Nexus 3.y.z only.

The CRD fields which could implement this:

  - `spec.automaticUpdate.enabled`: boolean field which lets the Operator know whether or not it should perform automatic updates. Ignored if `spec.image` is not empty and is different from the default value. Defaults to `true`.
  - `spec. automaticUpdate.nexusVersion`: this field should contain the Nexus image minor minor version we should stay in.  Ignored if `spec.automaticUpdate.enabled` is set to `false` or `spec.image` is not empty and is different from the default value. Must be informed if `spec.automaticUpdate.enabled` is set to `true`.

If these fields are valid and the Operator determines an update is necessary it will:

  - keep track of the current version in memory;
  - change the deployment's image accordingly;
  - if the rolling deployment of the new image fails, inform the user via the Nexus CR status and change back to the previously deployed version;
  -  if the rolling deployment is successful, inform the user via the Nexus CR status;

The operator must be able to check for updates periodically. This can be one by starting a goroutine which checks this from time to time, but updating the deployment is tricky. This goroutine must keep track of all latest minors and expose that with a data structure that optimizes queries (such as map). Once it detects a new version is available, it must update the relevant data structure and initiate reconciliation for all deployed Nexus CRs.

**Additional context**
N/A

Yes.

LCaparelli:
TODO:

- [x] Implement a mechanism to track latest tags
- [x] Implement a mechanism to keep track of an ongoing deployment
- [x] Implement a mechanism to raise events
- [x] Add an interface on the Nexus CRD to configure automatic updates
- [x] Implement unit tests
- [x] Document this bad boy

Fix #132 


pkg/controller/nexus/update/updates.go
```diff
@@ -0,0 +1,156 @@
...
+	defer tagsMutex.RUnlock()
+	tag, ok = latestMicros[minor]
+	return
+}
+
+// CheckUpdatesLoop checks if there is an update available every "delay" duration.
+// It also receives a client used to query the API server so it can start reconciliations for the controller when a newer tag for a given minor is found.
+func CheckUpdatesLoop(reconcileNexus reconcile.Reconciler, client client.Client, delay time.Duration) {
+	for {
+		time.Sleep(delay)
```
ricardozanini, line 64: I'm not sure if we should have a thread on going here since the reconcile loop will be called on t even if there's no change in the resources managed by the controllers. We could take an advantage from their timer instead of having ours.

We could continue having a go routine/new thread in the main reconciliation loop though, since this process should be async.

pkg/controller/nexus/update/updates.go
```diff
@@ -0,0 +1,156 @@
...
+			continue
+		}
+
+		if updated {
+			nexusList, err := getNexusList(client)
+			if err != nil && !errors.IsNotFound(err) {
+				log.Errorf("Unable to list deployed Nexus CRs: %v", err)
+			}
+			for _, deployedNexus := range nexusList.Items {
+				result, err := reconcileNexus.Reconcile(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: deployedNexus.Namespace, Name: deployedNexus.Name}})
```
ricardozanini, line 83: We shouldn't call the reconcile loop manually. We can just update the deployment resource and wait for the reconciliation to happen.

pkg/controller/nexus/update/updates.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+
+		updated, err := parseTagsAndUpdate(tags)
+		if err != nil {
+			log.Errorf(unableToCheckUpdatesFormat, err)
+			continue
+		}
+
+		if updated {
+			nexusList, err := getNexusList(client)
```
ricardozanini, line 78: We should be aware to the instance being requested, not all of them.

pkg/controller/nexus/update/updates.go
LCaparelli: Oh man, I had no idea that reconcile is called even if there is no change to the watched resources. This will simplify logic **a lot**. Here we only need to update the `latestMicros` map then. My idea is to call `GetLatest` from the Deployment Manager whenever creating the expected Deployment CR.

So this thread will run and from time to time check for updates and updating its internal map. Meanwhile, when a reconciliation begins, the Deployment Manager starts and defines the Deployment using the `GetLatest` function when it reaches the image definition. If it's different from the previous, it will update the Deployment and Kubernetes handles the rest.

pkg/controller/nexus/update/updates.go
LCaparelli: This part won't be necessary if the reconcile loop is called periodically no matter what. See the other conversation above :-)

pkg/controller/nexus/update/updates.go
LCaparelli: This part won't be necessary if the reconcile loop is called periodically no matter what. See the other conversation above :-)

pkg/controller/nexus/update/updates.go
LCaparelli: Oh, I think I misunderstood what you meant. I think you suggested calling the checker from the reconcile loop in order to fetch the tags only when a reconcile loop runs. Yeah, that's a good idea! Thanks for the suggestion

LCaparelli:
@ricardozanini this new commit contains **completely untested** code (matter of fact, it fails unit tests due to nil dereference). It needs more work to refine it, but before doing that I wanted to discuss this approach with you so we can decide if we want to follow this path.

I'll summarize what it's doing right now below. Assume automatic updates are enabled with a valid CR:

  - the `pkg/controller/nexus/update/updates.go` file has been renamed to `pkg/controller/nexus/update/tags.go`. Any caller that needs to fetch a tag from a specific minor is to call `GetLatest` from it. This basically keeps a map of latest micro release, like before, but now it will only fetch new tags if and only if: it has been more than 6 hours since the last refresh AND `GetLatest` has been called. This makes sure we only query tags when we need them and only if our cache is stale, simplifying the logic somewhat in comparison with the previous approach using a goroutine that sleeped most of the time. lazy++
  - every time a reconciliation begins, the Deployment Manager will run `GetLatest` to appropriately define the image for the deployment. If it it's different, the resulting Deployment will be different and it will be updated.
  - when this happens, `handleUpdateScenario` from `pkg/controller/nexus/nexus_controller.go` will determine if it is an update. If it is, it starts a separate goroutine to monitor the update.
  - this goroutine is `MonitorUpdate` from `pkg/controller/nexus/update/monitor.go`. It basically queries the API server for the Deployment readiness. It's an implementation robust to failures when performing requests to the API server with retries implemented with exponential backoff. If automatic updates are enabled, it will also attempt to rollback the update if it fails.

A specific concern I have: do we need exponential backoffs or are they already implemented by the client? In the reconcile loop, if any request to the API server fails we simply requeue the reconcile request, making it robust to failures. Here we don't have that luxury,  so if the client does not implement exponential backoff or some other mechanism we need to implement our own.

I wanted to know your opinion about how this was implemented before continuing, so let me know your thoughts! Feel free to critique, I never implemented this sort of feature, so maybe I'm not following the best path here (plus I get to learn from critique, it's very much welcome).

ricardozanini:
Hi! 

This seems the day to go, I particularly liked the idea to always fetch the minor update tag, if it's new we update and that's it.

> this goroutine is MonitorUpdate from pkg/controller/nexus/update/monitor.go. It basically queries the API server for the Deployment readiness. It's an implementation robust to failures when performing requests to the API server with retries implemented with exponential backoff. If automatic updates are enabled, it will also attempt to rollback the update if it fails.

We might not need such complex implementation. If the deployment fails, k8s will rollback to a previous stage. We only need to know if a rolling back took place to NOT try to update it again and fall into a error loop (setting the `autoUpdate` to false in the CR), and also flag the Nexus instance status about it, so we can safely continue reconcile with the newly CR. If the update fails:

1. The Deployment image will be rolled back to the previous one (actually, the deployment will point to a previous ReplicaSet with the old image)
2. We set the `nexus.status.update` to a message/structure informing that the update failed
3. We write to the Events API notifying the failure and reason
4. We turn `autoUpdate` to **false**, so we won't try to update without a manual intervention

If the user figure out the problem they can turn the update back on again and we would try another turn.

LCaparelli:
@ricardozanini Yeah, that approach makes sense to me. There is one small problem though: currently, our managers validate and set defaults to the CR that comes in via the reconciliation loop. It's at that time that the Deployment Manager writes to `spec.Image` and sets a tag in it. From there, it follows on through the flow I explained above and all should be well. Note that we never update the actual CR via the API server, so if the user left `spec.image` blank, the "canonical" CR will always be left blank, as we make no modifications to it on the user's behalf.

However, if a rollback is necessary, then we need to actually set `spec.image` to the tag which was working prior to the update. If we don't do that and the user left `spec.Image` blank, the Deployment Manager would actually set the tag to `latest`, which could lead to changing the minor and potentially leading to another failed deployment.

Now, I don't like this approach very much as it adds business logic related to the CR in a package which didn't need to be as tightly coupled with the details of how the Deployment Manager sets default values. While developing this feature it also came to my attention that whenever we need default values for a certain thing, we need to import the Managers' packages, which usually leads to an import cycle.

I feel this is a good time to refactor the code a bit and move any default setting/validation logic to a separate package (`pkg/controller/nexus/resource/validation`, for example). This way we have a single source of truth for defaults and validation which could be leveraged by the managers _and_ other packages (including API users) without leading to an import cycle.

Wdyt? If you agree, think we can squeeze this last thing into this release? I know it's a bit bloated already, but doing so would simplify the logic for this monitor and increase overall cohesion. It also doesn't seem to be too complicated to implement, mostly moving stuff that is already implemented elsewhere and introducing a function call or two to the Supervisor loop.

And as for the exponential backoff implementation, looking at it after a weekend and a nice cup of coffee, we can easily simplify by just retrying whatever operations are necessary inside the main loop of `MonitorUpdate` on `pkg/controller/nexus/update/monitor.go`:

```go
	for time.Since(updateStart) < readinessTimeout {
```

Basically log an unknown status if  any calls to the API server fail and try again on the next iteration, which should take place in another minute until a timeout is reached. Finally create an Event letting the user know whatever outcome we had and then ending of the monitor function.

ricardozanini:
Makes sense to me separating the validation from the manager package, being able to use these functions anywhere else. We can do it in this release for sure.

LCaparelli:
> Makes sense to me separating the validation from the manager package, being able to use these functions anywhere else. We can do it in this release for sure.

Ack, thanks for the prompt response! I'll work on this via #134 :grin:

LCaparelli:
@ricardozanini not sure why this could not be auto-merged, but it can be solved by accepting mine. If you take a closer look you'll see that mine only adds a couple of imports and actually keeps the one that triggered the conflict. I'd resolve myself, but it needs write access to the repo.

No rush in resolving though, the tests won't pass :P

ricardozanini:
It seems that we are failing in the e2e :(

LCaparelli:
Yeah, I changed a *LOT* of things. I was hoping it wouldn't break, but kind of expecting it. :-/ 

I'll take a closer look when I finish the unit tests.

README.md
```diff
@@ -96,6 +99,99 @@ Considering that you ran the install command above, to remove the operator compl
...
+Or you may query all events:
+
+```
+$ kubectl get events
+```
+
+A successful update event looks like:
+
+```yaml
+
```
ricardozanini, line 150: Where's the output? :D

pkg/controller/nexus/event/service.go
```diff
@@ -0,0 +1,66 @@
...
+//
+//     Nexus Operator is distributed in the hope that it will be useful,
+//     but WITHOUT ANY WARRANTY; without even the implied warranty of
+//     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//     GNU General Public License for more details.
+//
+//     You should have received a copy of the GNU General Public License
+//     along with Nexus Operator.  If not, see <https://www.gnu.org/licenses/>.
+
+package event
```
ricardozanini, line 18: I think this package could be moved to `pkg/client/kubernetes` instead, since there's no information regarding the controller itself.

pkg/controller/nexus/resource/validation/validation.go
```diff
@@ -184,6 +187,43 @@ func (v *Validator) setProbeDefaults(nexus *v1alpha1.Nexus) {
...
+			return
+		}
+		nexus.Spec.AutomaticUpdate.MinorVersion = &minor
+	}
+
+	log.Debugf("Fetching the latest micro from minor %d", *nexus.Spec.AutomaticUpdate.MinorVersion)
+	tag, ok := update.GetLatestMicro(*nexus.Spec.AutomaticUpdate.MinorVersion)
+	if !ok {
+		// the informed minor doesn't exist, let's just use 'latest'
+		log.Warnf("Latest tag for minor version (%d) not found. Using 'latest' instead and disabling Automatic", *nexus.Spec.AutomaticUpdate.MinorVersion)
```
ricardozanini, line 218: Why not fetching latest available and set `Spec.AutomaticUpdate.MinorVersion=nil` ?

ricardozanini: Also, I think we could add all those WARNs in form of Events instead.

pkg/controller/nexus/update/tags.go
```diff
@@ -0,0 +1,167 @@
...
+	return greatestMinor, nil
+}
+
+func fetchUpdates() {
+	tags, err := getTags()
+	if err != nil {
+		log.Errorf(unableToCheckUpdatesFormat, err)
+		return
+	}
+	lastQuery = time.Now()
```
ricardozanini, line 104: We might need an interval if we failed to fetch. Since we can fetch all day long uninterruptedly. I'd say 10 minutes and stop after 10 tries. Then try 6 hours later (using the `lastQuery`  ttl).

pkg/controller/nexus/update/monitor.go
```diff
@@ -0,0 +1,177 @@
+//     Copyright 2020 Nexus Operator and/or its authors
+//
+//     This file is part of Nexus Operator.
+//
+//     Nexus Operator is free software: you can redistribute it and/or modify
+//     it under the terms of the GNU General Public License as published by
```
ricardozanini, line 6: The headers must be updated to Apache.

README.md
LCaparelli: Oops! I think that's when I found the bug, got interrupted by it lol

I'll add it in a bit, thanks :-)

pkg/controller/nexus/event/service.go
LCaparelli: That's a good point! I think you meant `pkg/cluster/kubernetes` though, right?  My only objection is that keeping this in a separate package promotes re-usability by making it a small, concise package that manages events and only events. Wdyt of `pkg/cluster/event`?

pkg/controller/nexus/resource/validation/validation.go
LCaparelli: > Why not fetching latest available and set Spec.AutomaticUpdate.MinorVersion=nil ?

One of the objectives of this function is assigning a sane value to `minorVersion` if left uninformed. On line 203 we check for that and put in the latest available minor if left blank by the user.

In that branch of execution (when the user didn't specify a minor) we MUST have a valid map of minor->tag, otherwise we would have gotten a non-nil error on line 206:

```go
203     if nexus.Spec.AutomaticUpdate.MinorVersion == nil {
204         log.Debugf("Automatic Updates are enabled, but no minor was informed. Fetching the most recent...")
205         minor, err := update.GetLatestMinor()
206         if err != nil {
207             log.Errorf("Unable to fetch the most recent minor. Disabling automatic updates.")
208             nexus.Spec.AutomaticUpdate.Disabled = true
209             return
210         }
211         nexus.Spec.AutomaticUpdate.MinorVersion = &minor
212     }
```

So in that scenario `ok` must be equal to `true`.

The other scenario is if the user has specified a minor, but it has not been found. For the most part, hitting this means the CR simply doesn't have a valid minor, so going for the latest minor and keeping automatic updates enabled does make sense here. This can also mean that we couldn't fetch the tags, although the user has informed a valid minor. In that scenario we also would not have the tag for the latest minor, so we couldn't simply use that, meaning we would need to use `latest`.

Does that flow sound good to you @ricardozanini? Try the latest available minor while keeping automatic updates enabled. If that also isn't available, disable updates and use `latest`.

> Also, I think we could add all those WARNs in form of Events instead.

Well, that wouldn't be consistent with how we inform the user of other changes performed during validation. Are you suggesting we raise events for all changes performed by the validation package?

pkg/controller/nexus/update/tags.go
LCaparelli: Good point, I had not considered that. However, this would only be tried every reconcile, which happen automatically every hour or if there are changes to the watched objects. Is it really that bad if we attempt to fetch it on every reconcile?

If we do implement this mechanism to control how often we try again in case of failure, are you thinking of keeping it within the same reconcile iteration? Or are you thinking of trying again only on the next reconcile AND if we didn't fail in the last 10 minutes?

I assume you mean the latter, keeps a reconcile from possibly lasting 100 minutes, but just making sure we're on the same page.

After the 10th failure we can just do `lastQuery = time.Now()` and reset the failure count.

pkg/controller/nexus/update/monitor.go
LCaparelli: Ah, I thought `make test` would have done it. I guess I lost our bet then? Just switching the flag passed to `addlicense` did not replace existing headers. Bummer.

pkg/controller/nexus/event/service.go
ricardozanini: I'd like to call this like `kubernetes.RegisterEvent()` and not `event.RegisterEvent()` 😝

pkg/controller/nexus/update/monitor.go
ricardozanini: No, that's why I had to use regexp to remove all headers first. :(

pkg/controller/nexus/resource/validation/validation.go
ricardozanini: > Does that flow sound good to you @ricardozanini? Try the latest available minor while keeping automatic updates enabled. If that also isn't available, disable updates and use latest.

Makes sense.

>  Are you suggesting we raise events for all changes performed by the validation package?

We haven't the `event` feature at the point we created the validation package. So, at the time was OK using WARNs. Since now we have, we should inform the user via Events API/CR Status that something is not right. For this scenario I think add an Event would make sense and also have an update in the CR status.

pkg/controller/nexus/update/tags.go
ricardozanini: I think we can add a simple guard in this function in case the reconcile loop happens sequentially like 4/5 times within a short period. This would avoid keeping fetching the docker API. So I'd say to not fetch if it was an error in the last minute. This would do.

LCaparelli: Sounds like a reasonable middle ground, will do.

pkg/controller/nexus/resource/validation/validation.go
LCaparelli: I think a reasonable policy for raising events would be to raise one if it makes changes to the CR that are not simply default setting. This way the user gets warned on multiple fronts that their original CR being modified in ways they might not expect. Sounds good?

pkg/controller/nexus/event/service.go
LCaparelli: I'm confused, the functions we expose are `RaiseInfoEventf()` and `RaiseWarnEventf()`. They would be `event.RaiseInfoEventf()` and `event.RaiseWarnEventf()`. A bit repetitive, but I don't think it's very harmful... I think it's a worthy trade-off to keep package cohesion high. Or do you actually want to rename them to `RegisterEvent()`?

pkg/controller/nexus/resource/validation/validation.go
LCaparelli: But maybe this is better in another PR, with events for all situations in which this happens.

pkg/controller/nexus/event/service.go
ricardozanini: > A bit repetitive, but I don't think it's very harmful

Event is a Kubernetes API call, that's why I think would make much more sense then `event` package. We don't want to keep creating too many packages.

pkg/controller/nexus/resource/validation/validation.go
ricardozanini: Yes in another PR would be better, but for this particular scenario in this comment, I think we can change to the event instead since it seems pretty important to inform the user what happened. wdyt?

pkg/controller/nexus/event/service.go
LCaparelli: Ok, we'll keep it in `pkg/cluster/kubernetes`

pkg/controller/nexus/resource/validation/validation.go
LCaparelli: Will do

LCaparelli:
@ricardozanini I made the changes, but I just realized this is getting into:

`API rule violation: list_type_missing,./pkg/apis/apps/v1alpha1,NexusStatus,UpdateConditions`

when generating the CSV. Now, I'm not sure how to tackle this, since `Spec.Status.UpdateConditions` is a `[]string`, why would it need a list type?

Any ideas?

go.mod
```diff
@@ -25,12 +26,9 @@ require (
 )
 
 replace (
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
 	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20190814194116-a94e914914f4+incompatible
-	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20190813201236-5a5508328169
```
ricardozanini, line 29: We don't need OpenShift's client any more?

deploy/olm-catalog/nexus-operator/0.3.0/nexus-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -62,6 +62,9 @@ spec:
         name: A Kubernetes Service
         version: v1
       specDescriptors:
+      - description: Automatic updates configuration
```
ricardozanini, line 65: Could you please also change the capability level to 2? :sunglasses:

go.mod
LCaparelli: The IDE removed it on its own, so I guess not? It's compiling alright. Should we check anything else?

`grep -r 'github.com/openshift/client-go' *` returns nothing as well.

deploy/olm-catalog/nexus-operator/0.3.0/nexus-operator.v0.3.0.clusterserviceversion.yaml
LCaparelli: Hell yeah :sunglasses:

go.mod
ricardozanini: Removing dependencies is good. :sunglasses:

PR #135, opened 2020-08-03
Move default/validation logic to separate package

### LINKED ISSUE #134: Move CR default setting/validation logic to a separate package
Currently logic for deaults and validation is distributed to each resource manager. This makes it difficult to use default values without creating import cycles and it complicates maintenance as there are several sources of logic which need to be taken into account when making changes.

If we keep all logic related to defaults and validation in a single package (`pkg/controller/nexus/resource/validation`, for example) we solve both of these issues.

It would also be nice if this package updated the CR on the API server after all operations are performed, thus making sure the desired state stored in the control plane is valid and can be easily consulted at any time. This also makes sure that automatic changes to the CR are always visible to users, opposed to the implicit approach we have today, where the CR provided by the user is never updated, providing no feedback with regards to the defaults which were set.

LCaparelli:
Previously logic for defaults and validation was distributed to each
resource manager. This makes it difficult to use default values without
creating import cycles and it complicates maintenance as there are
several sources of logic which need to be taken into account when making
changes.

If we keep all logic related to defaults and validation in a single
package we solve both of these issues.

This commit also makes sure the CR gets updated in the API server after
defaults were set and it has been validated, thus making sure the
desired state stored in the control plane is valid and can be easily
consulted at any time.

Fix #134


LCaparelli:
Huh. Something wrong with codecov, but I didn't really change anything there.

```
->  Pinging Codecov
https://codecov.io/upload/v4?package=bash-20200728-9fb7d93&token=secret&branch=&commit=88034d45d79b9b7f67bfcceb564c503139c05a0e&build=&build_url=&name=operator-test&tag=&slug=m88i%2Fnexus-operator&service=&flags=operator&pr=&job=&cmd_args=f,n,F,Z
HTTP 400
Please provide the repository token to upload reports via `-t :repository-token`
```

I'll see what I can dig up.

ricardozanini:
Waiting for this to be fixed: https://github.com/actions/runner/issues/635

LCaparelli:
@ricardozanini good for merging :-)

PR #138, opened 2020-08-11
Add custom comparator for Deployments

### LINKED ISSUE #137: Deployments are deemed different even if there are no changes in the Nexus CR
**Describe the bug**
Deployments are deemed different even if there are no changes in the Nexus CR. This is due to the server setting default values we don't touch when creating the Deployment and the default comparator running `reflect.DeepEqual()` on the Deployment spec.

**To Reproduce**
1. Produce a CR which does not set the image Pull Policy
2. Observe the logs for the second reconciliation

**Expected behavior**
Required and deployed Deployments are deemed equal and it doesn't get updated.

**Environment**
Current version, any platform.

**Additional context**
We can bypass this altogether by having our own custom comparator that only cares about what we change. This way we don't have to worry about defaults the API server would set anyway and we won't perform as many comparisons using reflection, which will likely improve performance. On the other hand it's more code to maintain, but probably worth the trouble.

LCaparelli:
Fix #137

### DIFF
--- a/pkg/controller/nexus/resource/deployment/deployment.go
+++ b/pkg/controller/nexus/resource/deployment/deployment.go
@@ -90,9 +90,6 @@ func newDeployment(nexus *v1alpha1.Nexus) *appsv1.Deployment {
 					},
 				},
 			},
-			Strategy: appsv1.DeploymentStrategy{
-				Type: appsv1.RecreateDeploymentStrategyType,
-			},
 		},
 	}
 
--- a/pkg/controller/nexus/resource/deployment/manager.go
+++ b/pkg/controller/nexus/resource/deployment/manager.go
@@ -21,6 +21,7 @@ import (
 	ctx "context"
 	"fmt"
 	"github.com/RHsyseng/operator-utils/pkg/resource"
+	"github.com/RHsyseng/operator-utils/pkg/resource/compare"
 	"github.com/m88i/nexus-operator/pkg/apis/apps/v1alpha1"
 	"github.com/m88i/nexus-operator/pkg/logger"
 	appsv1 "k8s.io/api/apps/v1"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 )
 
 var log = logger.GetLogger("deployment_manager")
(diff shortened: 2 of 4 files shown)


LCaparelli:
It seems it's failing to validate the codecov yaml as it cannot connect. Running fine on my system:

```
╰─ curl --data-binary @codecov.yml https://codecov.io/validate     
Valid!

{
  "flags": {
    "operator": {
      "paths": [
        "^pkg/.*"
      ]
    }
  },
  "coverage": {
    "status": {
      "project": {
        "operator": {
          "target": "auto",
          "threshold": 5.0,
          "base": "auto",
          "flags": [
            "operator"
          ]
        }
      }
    }
  },
  "comment": {
    "require_changes": true
  }
}
```

We probably shouldn't fail the CI because of this anyway... I'll do some digging.

Kaitou786:
> LGTM. I'm just curious about separating the coverage tasks in the CI. We are uploading, downloading and sending it again.
> 
> Since I've also added the coverage for e2e on #120 , not sure how we could merge both reports to increase our coverage percentages. @Kaitou786 do you know?

I'm not sure but I think this could be done, I'll do some search about it and let you guys know.
 Also, I'm not sure why we are uploading downloading and sending report again :/

LCaparelli:
@ricardozanini @Kaitou786 the CI was failing because the [validation of the codecov config was failing](https://github.com/m88i/nexus-operator/pull/138/checks?sha=f6fb3b865eee457e17b04afa722b4bb84fe6a105) and this led to the failure of the whole CI. It prevented integration tests from running even though unit tests and linting was successful. 

IMO we shouldn't fail our CI these tasks fail, they are secondary.

I tried finding a way to not fail the CI if these steps failed, but Github Actions only offer that for [entire jobs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), not steps from a job. And even that would have [unwanted outcomes](https://github.community/t/continue-on-error-allow-failure-ui-indication/16773) as the job would be considered successful if any steps failed, including the unit tests themselves.

The best way I found to prevent these tasks from failing our CI while keeping consistent outcomes was putting it in a separate job. I'm not convinced this hinders performance in a noticeable manner and we get the benefit of removing codecov tasks as dependency for our integration tests.

Not to say that's the best way, if you guys find something better let me know :grin:

ricardozanini:
But after your change unfortunately the codecov check stops working. :(

LCaparelli:
Oh shoot, I hadn't realized that actually. And to be honest, I'm not sure why it stopped working... :-(

I'll do some digging.

LCaparelli:
@ricardozanini @Kaitou786 I didn't find a solution for this with github actions, but I managed to hack something up to prevent the validation from failing the whole CI if it fails. `curl -vvv --data-binary @codecov.yml https://codecov.io/validate || exit 0` will make it always exit 0 and succeed the step. It sort of defeats the purpose of having the check though, we'd need to check it even if it's successful as it might have actually failed.

Ideally github would allow us to fail the step without failing the whole CI, but I didn't find a way to make this happen.

I was going to say I'm happy with just reverting the commits on which I changed this and going with what we had before, but our validation step just [failed again](https://github.com/m88i/nexus-operator/pull/138/checks?check_run_id=976004404) when attempting to connect to codecov.io... Apparently there is some outage going back and forth there (or maybe at github? the request might not be leaving GH's network).

I'm thinking about reverting the commits anyway and force push until the validation succeeds as it seems to be intermittent.

ricardozanini:
You can just revert and then when it fails, go the the GH actions tab and click "run job again" no need to darth vader all the time.

LCaparelli:
> You can just revert and then when it fails, go the the GH actions tab and click "run job again" no need to darth vader all the time.

LOL
e87fd270d69f2a4a74324de14e7c41161dcefb0ar1-480-204_hq

I must be blind, but I never spotted that button. Is it only available if the run fails?

ricardozanini:
> I must be blind, but I never spotted that button. Is it only available if the run fails?

Always there! Or maybe you don't have the right permissions?

LCaparelli:
@Kaitou786 when you got some bandwitdth would you mind reviewing the changes? :-)

Kaitou786:
> @Kaitou786 when you got some bandwitdth would you mind reviewing the changes? :-)

I looked at the code yesterday, forgot to approve. sorry about the delay :/

LCaparelli:
> I looked at the code yesterday, forgot to approve. sorry about the delay :/

It's all good, no worries :-D

PR #145, opened 2020-08-27
Change references to 'main' branch

LCaparelli:
**Note**: easy to miss, but this PR is to merge to 'main', not 'master'.

Change references of the 'master' branch to the 'main' branch in our files, except for previous versions CSVs. The branch itself has already been created, but it needs to be set as default when we're ready to migrate.

I wonder if we should keep the 'master' branch alive so that links of previous releases pointing to our README don't break. We could add a banner at the top of the master's README explaining it's only kept for historical reasons and link it to the main branch.

Alternatively we could update previous releases links as well, but we'd need to push them again to Operator Hub. This would allow us to delete master without any issues. This would still break any other links to our repo that specifically use the master branch, though these aren't maintained by us and there isn't much to be done if we chose to delete the master.

### DIFF
--- a/.github/workflows/nexus-operator-integration-checks.yml
+++ b/.github/workflows/nexus-operator-integration-checks.yml
@@ -10,7 +10,7 @@ on:
       - "Makefile"
       - ".github/ISSUE_TEMPLATE/**"
     branches:
-      - master
+      - main
 env:
   OPERATOR_SDK_VERSION: v0.18.1
   GO_VERSION: 1.14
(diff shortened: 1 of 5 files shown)


LCaparelli:
That said, since we're small we probably can get away with deleting the master branch and breaking 3rd party links to us without causing too much trouble, if any at all.

ricardozanini:
I'd say that as soon as we release 0.3 at OperatorHub we can get rid of the `master` branch. I'll merge this now after CI and  put main as the default one.

Issue #154, opened 2020-09-20
Allow remote installation


LCaparelli:
Today users who would like to install the Operator without OLM would use the steps described in our [quick install guide](https://github.com/m88i/nexus-operator#quick-install). While it works great, the user must have the project's files locally, as it uses our Makefile which in turn runs the actual [installation script](https://github.com/m88i/nexus-operator/blob/main/hack/install.sh) (which *also* requires locally downloaded files).

`kubectl` is capable of fetching resources using HTTP, like so:

  `$ kubectl apply -f https://raw.githubusercontent.com/m88i/nexus-operator/main/deploy/operator.yaml`

And yaml syntax allows having several yaml in the same file, meaning we could compile `deploy/operator.yaml`, `deploy/role_binding.yaml`, `deploy/role.yaml` and `deploy/service_account.yaml` into a single file and then simply apply it, thus rendering `hack/install.sh` obsolete.

Installing it remotely would be as simple as:

  `$ kubectl apply -f https://raw.githubusercontent.com/m88i/nexus-operator/<some path>/operator.yaml`

The [README quick install section](https://github.com/m88i/nexus-operator#quick-install) would also need to be updated to reflect the new instructions.

PR #159, opened 2020-09-24
Move CSV to 0.4.0

### LINKED ISSUE #151: Correct mention of Ingress Controllers on CSV
We mention we deploy an Ingress Controller in our CSV:

```
╰─ grep -i "ingress controller" deploy/olm-catalog/nexus-operator/0.3.0/nexus-operator.v0.3.0.clusterserviceversion.yaml 
    * Creates an [Ingress controller](https://kubernetes.io/docs/concepts/services-networking/ingress/) in Kubernetes (1.14+) environments to expose the application to the world
```

However we really only deploy ingresses, no ingress controllers. This needs to be changed on the next release's CSV.

LCaparelli:
Fix #151

### DIFF
--- a/deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
@@ -134,8 +134,10 @@ spec:
     Creates a new Nexus 3.x deployment in a Kubernetes cluster. Will help DevOps to have a quick Nexus application exposed to the world that can be used in a CI/CD process:
 
     * Deploys a new Nexus 3.x server based on either Community or Red Hat images
-    * Creates an [Ingress controller](https://kubernetes.io/docs/concepts/services-networking/ingress/) in Kubernetes (1.14+) environments to expose the application to the world
+    * Creates an [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) in Kubernetes (1.14+) environments to expose the application to the world
     * On OpenShift, creates a Route to expose the service outside the cluster
+    * Automatically creates Apache, Red Hat and JBoss Maven repositories
+    * Automatically updates Nexus within the same minor version
 
     [See our documentation](https://github.com/m88i/nexus-operator/blob/main/README.md) for more installation and usage scenarios.
 
--- a/deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
@@ -35,7 +35,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Developer Tools
     certified: "false"
-    containerImage: quay.io/m88i/nexus-operator:0.3.0
+    containerImage: quay.io/m88i/nexus-operator:0.4.0
     createdAt: "2019-11-16T13:12:22Z"
     description: Nexus Operator to deploy and manage Nexus 3.x servers
     repository: https://github.com/m88i/nexus-operator
@@ -177,7 +177,7 @@ spec:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
                   value: nexus-operator
-                image: quay.io/m88i/nexus-operator:0.3.0
+                image: quay.io/m88i/nexus-operator:0.4.0
                 imagePullPolicy: Always
                 name: nexus-operator
                 resources: {}
--- a/deploy/olm-catalog/nexus-operator/manifests/nexus-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/nexus-operator/manifests/nexus-operator.clusterserviceversion.yaml
@@ -134,8 +134,10 @@ spec:
     Creates a new Nexus 3.x deployment in a Kubernetes cluster. Will help DevOps to have a quick Nexus application exposed to the world that can be used in a CI/CD process:
 
     * Deploys a new Nexus 3.x server based on either Community or Red Hat images
-    * Creates an [Ingress controller](https://kubernetes.io/docs/concepts/services-networking/ingress/) in Kubernetes (1.14+) environments to expose the application to the world
+    * Creates an [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) in Kubernetes (1.14+) environments to expose the application to the world
     * On OpenShift, creates a Route to expose the service outside the cluster
+    * Automatically creates Apache, Red Hat and JBoss Maven repositories
+    * Automatically updates Nexus within the same minor version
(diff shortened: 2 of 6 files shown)


LCaparelli:
Oops, something's wrong. Looking

LCaparelli:
Not sure what's going on, the e2e tests ran fine locally. At the CI it just times out waiting for the first test's deployment. Not to mention I didn't even change anything related to the e2e tests, so this seems pretty weird. Any ideas?

Kaitou786:
Should the image tag in `deploy/operator.yaml` be changed to 0.4?

ricardozanini:
> Should the image tag in `deploy/operator.yaml` be changed to 0.4?

yes it should

deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,325 @@
...
+              }
+            },
+            "useRedHatImage": false
+          }
+        }
+      ]
+    capabilities: Seamless Upgrades
+    categories: Developer Tools
+    certified: "false"
+    containerImage: quay.io/m88i/nexus-operator:0.3.0
```
ricardozanini, line 38:
```suggestion
    containerImage: quay.io/m88i/nexus-operator:0.4.0
```

deploy/olm-catalog/nexus-operator/0.4.0/nexus-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,325 @@
...
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.annotations['olm.targetNamespaces']
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: OPERATOR_NAME
+                  value: nexus-operator
+                image: quay.io/m88i/nexus-operator:0.3.0
```
ricardozanini, line 180:
```suggestion
                image: quay.io/m88i/nexus-operator:0.4.0
```

deploy/olm-catalog/nexus-operator/manifests/nexus-operator.clusterserviceversion.yaml
```diff
@@ -41,7 +41,7 @@ metadata:
     repository: https://github.com/m88i/nexus-operator
     support: m88i Labs
     tectonic-visibility: ocs
-  name: nexus-operator.v0.3.0
+  name: nexus-operator.v0.4.0
```
ricardozanini, line 44: The image name should be changed as well :P

ricardozanini (changes requested):
Just update the image version and we should be ready to go.

LCaparelli:
> Should the image tag in deploy/operator.yaml be changed to 0.4?

>> yes it should

If we change it users will need to build the 0.4 image locally since there's no 0.4 on Quay, no? I think we should only change it on the next release, especially considering `deploy/operator.yaml` will be used to install the operator remotely without cloning the repo in #154

But I wonder if this has to do with the failure of the e2e tests, since on previous releases `deploy/operator.yaml` was indeed pointed to the next release, not the current one IIRC. If this is the case, we might want to consider having an `operator.yaml` file as one of the assets per release, this way we freeze it per release with a readable URL (as opposed to one pointing to a specific commit hash).

But given we change PullPolicy on our e2e script to `Never` before running, it still would just use whatever image it just built, regardless of its tag, no?

```
 37     echo "---> Updating deployment file to imagePullPolicy: Never"
 38     sed -i.bak 's/imagePullPolicy:\s*Always/imagePullPolicy: Never/g' ./deploy/operator.yaml
```

I even tried building with Alpine:latest locally as base image, but no luck.

> Just update the image version and we should be ready to go.

Yeah, I completely missed these last night, my bad.

Kaitou786:
> If we change it users will need to build the 0.4 image locally since there's no 0.4 on Quay, no? I think we should only change it on the next release, especially considering deploy/operator.yaml will be used to install the operator remotely without cloning the repo in #154

I think we can keep it with the next release version, 0.4 . For #154  we can have a similar procedure we've with the `make install`, i.e. checkout the repository at the latest tag(0.3), and then apply the `operator.yaml`

> But given we change PullPolicy on our e2e script to Never before running, it still would just use whatever image it just built, regardless of its tag, no?

Could be the image is build with 0.4 tag which the deployment is not able to find, hence the deployment not running?

LCaparelli:
> I think we can keep it with the next release version, 0.4 . For #154 we can have a similar procedure we've with the make install, i.e. checkout the repository at the latest tag(0.3), and then apply the operator.yaml

The idea there is to have users run `kubectl apply -f https://bla-bla/operator.yaml` in order to install everything. We first considered a similar remote installation script which would look a lot like what `make install` does, but ultimately just running `kubectl apply <...>`is just simpler and I think we should keep it that way. 

> Could be the image is build with 0.4 tag which the deployment is not able to find, hence the deployment not running?

Ah yes, that makes sense, since the tests are performed by applying the yaml files `deploy` (which points to 0.3, not 0.4). It's probably working on mine because I have 0.3 cached as I didn't rebuild my minikube VM after some tests I was performing with 0.3. Nice one @Kaitou786, thanks! :-D

@ricardozanini @Kaitou786 I'll change `deploy/operator.yaml` for now, but let's continue this discussion on #154 or telegram regarding how we should allow remote installation.

Kaitou786:
Ah, yes @LCaparelli I was thinking if we are going to maintain a separate file with everything in it, the image tag there can be for the last released version of our operator. Just a though :p. We can discuss this more :)

LCaparelli:
> Ah, yes @LCaparelli I was thinking if we are going to maintain a separate file with everything in it, the image tag there can be for the last released version of our operator.

Absolutely, that was one of the approaches I was thinking about too. We just have to keep it off the `deploy` directory due to this (I don't think that's a bad thing, just pointing out the pitfall I just fell in hehe).

LCaparelli:
@ricardozanini @Kaitou786 CI passing now :-)

Kaitou786 (approved):
Great!

ricardozanini:
@LCaparelli the idea is to not have a separate operator.yaml file since it's alone do nothing. We need RBAC as well.. so the idea is to combine everything in a single file like `nexus-operator.yaml` and make it part of the release.

So users would simply `kubectl apply -f <github_release_url>/0.4.0/nexus-operator.yaml` to install it.

ricardozanini (approved):

LCaparelli:
> @LCaparelli the idea is to not have a separate operator.yaml file since it's alone do nothing. We need RBAC as well.. so the idea is to combine everything in a single file like `nexus-operator.yaml` and make it part of the release.
> 
> So users would simply `kubectl apply -f <github_release_url>/0.4.0/nexus-operator.yaml` to install it.

@ricardozanini Yes, I understand that. I mention that on the issue hehe :-)

Issue #158, opened 2020-09-23
Tighten up the Role permissions


LCaparelli:
Currently our role has access to a lot of resources it doesn't really manipulate as a result of previous problems which no longer exist. We should "clean" it up, reducing the surface area of operations allowed for our service account and thus improving security.

Currently we have:

```
╰─ kubectl describe role/nexus-operator
Name:                                    nexus-operator
Labels:                                  <none>
Annotations:                             PolicyRule:
  Resources                              Non-Resource URLs  Resource Names    Verbs
  ---------                              -----------------  --------------    -----
  pods                                   []                 []                [create delete get list patch update watch get]
  deployments.apps                       []                 []                [create delete get list patch update watch get]
  replicasets.apps                       []                 []                [create delete get list patch update watch get]
  configmaps                             []                 []                [create delete get list patch update watch]
  endpoints                              []                 []                [create delete get list patch update watch]
  events                                 []                 []                [create delete get list patch update watch]
  persistentvolumeclaims                 []                 []                [create delete get list patch update watch]
  podtemplates                           []                 []                [create delete get list patch update watch]
  replicationcontrollers                 []                 []                [create delete get list patch update watch]
  secrets                                []                 []                [create delete get list patch update watch]
  serviceaccounts                        []                 []                [create delete get list patch update watch]
  services/finalizers                    []                 []                [create delete get list patch update watch]
  services                               []                 []                [create delete get list patch update watch]
  *.apps.m88i.io                         []                 []                [create delete get list patch update watch]
  daemonsets.apps                        []                 []                [create delete get list patch update watch]
  statefulsets.apps                      []                 []                [create delete get list patch update watch]
  ingresses.networking.k8s.io            []                 []                [create delete get list patch update watch]
  routes.route.openshift.io              []                 []                [create delete get list patch update watch]
  servicemonitors.monitoring.coreos.com  []                 []                [get create]
  deployments.apps/finalizers            []                 [nexus-operator]  [update]
```

LCaparelli:
Dug around a bit today. I kept all resources we manipulate directly with all permissions and removed the others. Then I added permissions back as necessary while running all of our examples on minikube and minishift, ended up with this:

```
  Resources                              Non-Resource URLs  Resource Names    Verbs
  ---------                              -----------------  --------------    -----
  events                                 []                 []                [create delete get list patch update watch]
  persistentvolumeclaims                 []                 []                [create delete get list patch update watch]
[...]

PR #162, opened 2020-09-26
Fix #156, equalizing SecurityContext object

### LINKED ISSUE #156: When deploying Nexus with Red Hat image on OpenShift, deployment get stuck in the operator reconciliation loop
**Describe the bug**
When deploying a Nexus server with the operator on OpenShift using Red Hat image, the operator enters in a infinite reconciliation loop:

```json
{"level":"info","T":"2020-09-22T21:04:47.309Z","logger":"controller_nexus","msg":"Reconciling Nexus 'nexus3' on namespace 'nexus'"}
{"level":"warn","T":"2020-09-22T21:04:47.342Z","logger":"Nexus_validation","msg":"Nexus CR configured to the use Red Hat Certified Image, ignoring 'spec.image' field."}
{"level":"info","T":"2020-09-22T21:04:47.379Z","logger":"resources_management","msg":"Fetching required resources"}
{"level":"info","T":"2020-09-22T21:04:47.379Z","logger":"resources_management","msg":"Fetching deployed resources"}
{"level":"info","ts":1600808687.380765,"logger":"comparator","msg":"Objects are not equal","deployed":"&PodSecurityContext{SELinuxOptions:nil,RunAsUser:nil,RunAsNonRoot:nil,SupplementalGroups:[],FSGroup:nil,RunAsGroup:nil,Sysctls:[]Sysctl{},WindowsOptions:nil,FSGroupChangePolicy:nil,}","requested":"nil"}
{"level":"info","T":"2020-09-22T21:04:47.380Z","logger":"controller_nexus","msg":"Will create 0, update 1, delete 0 instances of v1.Deployment"}
{"level":"info","T":"2020-09-22T21:04:47.401Z","logger":"controller_nexus","msg":"Performing Nexus server operations if needed"}
{"level":"info","T":"2020-09-22T21:04:47.401Z","logger":"controller_nexus","msg":"Server Operations finished. Status is {false false false false Server does not have enough availble replicas}"}
{"level":"info","T":"2020-09-22T21:04:47.401Z","logger":"controller_nexus","msg":"Updating application status before leaving"}
{"level":"info","T":"2020-09-22T21:04:47.401Z","logger":"controller_nexus","msg":"Checking Deployment Status"}
{"level":"info","T":"2020-09-22T21:04:47.401Z","logger":"controller_nexus","msg":"Checking Route Status"}
{"level":"info","T":"2020-09-22T21:04:47.406Z","logger":"controller_nexus","msg":"Updating status for Nexus instance 'nexus3'"}
{"level":"info","T":"2020-09-22T21:04:48.448Z","logger":"controller_nexus","msg":"Controller finished reconciliation"}
```

This is because the `PodSecurityContext` comparison is failing.

**To Reproduce**
Install the default Nexus CR in an OpenShift cluster with the attribute `useRedHatImage: true` and no persistence.

**Expected behavior**
The Nexus server to be deployed.

**Environment**
CRC 4.5

ricardozanini:
This PR fixes #156 

@LCaparelli also tested the new RBAC rules from #158 and it's working perfectly.

Testing this changes, I came up with another bug related to this one: [Red Hat images with volatile configurations also requires the SCC](https://github.com/sonatype/docker-nexus3/blob/master/Dockerfile.rh.el#L73). The only configuration we won't need the SCC configured, is Red Hat images with a PV associated. That's because OpenShift will configure the mounted PV to be accessible by the [UID created in the image](https://github.com/sonatype/docker-rhel-nexus/blob/master/uid_entrypoint). Also see: https://github.com/sonatype/docker-nexus3#red-hat-certified-image

Signed-off-by: Ricardo Zanini <zanini@redhat.com>

### DIFF
--- a/pkg/controller/nexus/resource/deployment/manager.go
+++ b/pkg/controller/nexus/resource/deployment/manager.go
@@ -91,7 +91,7 @@ func deploymentEqual(deployed resource.KubernetesResource, requested resource.Ku
 	depDeployment := deployed.(*appsv1.Deployment)
 	reqDeployment := requested.(*appsv1.Deployment)
 
-	equalizeSecurityContext(depDeployment, reqDeployment)
+	normalizeSecurityContext(depDeployment, reqDeployment)
 
 	var pairs [][2]interface{}
 	pairs = append(pairs, [2]interface{}{depDeployment.Name, reqDeployment.Name})
@@ -132,7 +132,7 @@ func equalPullPolicies(depDeployment, reqDeployment *appsv1.Deployment) bool {
 // see: https://github.com/m88i/nexus-operator/issues/156
 // On OpenShift 4.5+ `SecurityContext` is not nil, but a "blank" object.
 // Since we are requesting a nil object in this context, we consider the deployed object to be nil as well.
-func equalizeSecurityContext(depDeployment, reqDeployment *appsv1.Deployment) {
+func normalizeSecurityContext(depDeployment, reqDeployment *appsv1.Deployment) {
 	if reqDeployment.Spec.Template.Spec.SecurityContext == nil {
 		depDeployment.Spec.Template.Spec.SecurityContext = nil
 	}
--- a/pkg/controller/nexus/resource/deployment/manager.go
+++ b/pkg/controller/nexus/resource/deployment/manager.go
@@ -90,6 +90,9 @@ func (m *Manager) GetCustomComparators() map[reflect.Type]func(deployed resource
 func deploymentEqual(deployed resource.KubernetesResource, requested resource.KubernetesResource) bool {
 	depDeployment := deployed.(*appsv1.Deployment)
 	reqDeployment := requested.(*appsv1.Deployment)
+
+	equalizeSecurityContext(depDeployment, reqDeployment)
+
 	var pairs [][2]interface{}
 	pairs = append(pairs, [2]interface{}{depDeployment.Name, reqDeployment.Name})
 	pairs = append(pairs, [2]interface{}{depDeployment.Namespace, reqDeployment.Namespace})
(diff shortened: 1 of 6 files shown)


Kaitou786 (approved):

pkg/controller/nexus/resource/deployment/manager.go
```diff
@@ -125,3 +128,12 @@ func equalPullPolicies(depDeployment, reqDeployment *appsv1.Deployment) bool {
...
 }
+
+// see: https://github.com/m88i/nexus-operator/issues/156
+// On OpenShift 4.5+ `SecurityContext` is not nil, but a "blank" object.
+// Since we are requesting a nil object in this context, we consider the deployed object to be nil as well.
+func equalizeSecurityContext(depDeployment, reqDeployment *appsv1.Deployment) {
+	if reqDeployment.Spec.Template.Spec.SecurityContext == nil {
+		depDeployment.Spec.Template.Spec.SecurityContext = nil
+	}
+}
```
LCaparelli, line 139: I got a couple of observations about this one.

  1. Once I read the function I understood what's going on, but when I saw its usage in line 94 I didn't understand if this was comparing two SCCs (didn't seem to be that since the parameters are deployments) or what. My understanding is that this ["normalizes" or "canonicalizes"](https://en.wikipedia.org/wiki/Canonicalization)  both possible representations of the SCC withing a deployment, how about naming it that way? `normalizeSecurityContext` or equivalent.
  2. Receiving the deployments makes for additional context, but a more verbose function. If we received the `*SecurityContext` directly, it could be a bit cleaner:

```suggestion
func normalizeSecurityContext(depCtx, reqCtx *corev1.SecurityContext) {
	if reqCtx == nil {
		depCtx = nil
	}
}
```
  Wdyt?

LCaparelli (commented):
Looks good to me! Thanks for this and for testing the new Role :-D

Left a couple of nitpicks, let me know your thoughts!

pkg/controller/nexus/resource/deployment/manager.go
ricardozanini: I won't change the method signature since we have two `SecurityContext` in the deployment spec. Might cause confusion to the caller. Let's the function decide with one (or ones maybe in the future) to normalize. I wrote the function's name with other thing in mind actually, that's why `equalize`.

PR #164, opened 2020-10-01
Fail CI if change in autogenerated files are detected

Kaitou786:


Kaitou786:
There were some auto-generated files changes needed to be committed to our main branch, I saw when the CI failed, so for that, I created #165. This commit, for now, have been rebased with #165 for verification, but those files from this PR will be removed when we merge that commit

hack/go-vet.sh
```diff
@@ -19,10 +19,10 @@ GOPATH=$(go env GOPATH)
 ./hack/addheaders.sh
 
 operator-sdk generate k8s
-operator-sdk generate crds
+operator-sdk generate crds --crd-version=v1beta1
```
ricardozanini, line 22: :clap:

ricardozanini (approved):

hack/go-vet.sh
LCaparelli: nice catch o/

hack/go-vet.sh
```diff
@@ -19,10 +19,10 @@ GOPATH=$(go env GOPATH)
...
 
 # get the openapi binary
-command -v openapi-gen >/dev/null || go build -o $GOPATH/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen
+command -v openapi-gen >/dev/null || go build -o $GOPATH/bin/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen
```
LCaparelli, line 25: Nice catch v2

LCaparelli (approved):

PR #170, opened 2020-10-06
Implement identifiable logging

### LINKED ISSUE #143: Make logs prefixed with the namespace and name of the Nexus being reconciled
It gets confusing when multiple Nexus instances are being managed in the same cluster or namespace.

SSIA

Modifying the logger to hold the name of the Nexus instance. When it's nil, don't add the prefix.

**Additional context**
N/A

Yes

LCaparelli:
As the operator moves towards being cluster-scoped it's important that
users are able to quickly identify context from logs and can tell
exactly which Nexus CR is the subject of logs. To that effect this
change adds the "namespace" and "resource name" fields to structured
logging when applicable.

Additionally, standard output is now the default log output and sync
operations were removed as they're not necessary due to stdout's
unbuffered nature.

Fix #143

### DIFF
--- a/pkg/controller/nexus/resource/networking/manager.go
+++ b/pkg/controller/nexus/resource/networking/manager.go
@@ -178,5 +178,9 @@ func ingressEqual(deployed resource.KubernetesResource, requested resource.Kuber
 	pairs = append(pairs, [2]interface{}{ingress1.Namespace, ingress2.Namespace})
 	pairs = append(pairs, [2]interface{}{ingress1.Spec, ingress2.Spec})
 
-	return compare.EqualPairs(pairs)
+	equal := compare.EqualPairs(pairs)
+	if !equal {
+		logger.GetLogger("networking_manager").Info("Resources are not equal", "deployed", deployed, "requested", requested)
+	}
+	return equal
 }
--- a/pkg/controller/nexus/resource/networking/manager.go
+++ b/pkg/controller/nexus/resource/networking/manager.go
@@ -16,21 +16,22 @@ package networking
 
 import (
 	"fmt"
-	"github.com/m88i/nexus-operator/pkg/framework"
-	"go.uber.org/zap"
 	"reflect"
 
 	"github.com/RHsyseng/operator-utils/pkg/resource"
 	"github.com/RHsyseng/operator-utils/pkg/resource/compare"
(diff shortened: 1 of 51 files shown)


pkg/controller/nexus/resource/networking/manager.go
```diff
@@ -204,9 +178,5 @@ func ingressEqual(deployed resource.KubernetesResource, requested resource.Kuber
 	pairs = append(pairs, [2]interface{}{ingress1.Namespace, ingress2.Namespace})
 	pairs = append(pairs, [2]interface{}{ingress1.Spec, ingress2.Spec})
 
-	equal := compare.EqualPairs(pairs)
-	if !equal {
-		log.Info("Resources are not equal", "deployed", deployed, "requested", requested)
```
ricardozanini, line 209: This log is pretty useful. Are you sure you want to remove it?

pkg/controller/nexus/server/users.go
```diff
@@ -106,15 +106,15 @@ func (u *userOperation) createOperatorUserIfNotExists() (*nexus.User, error) {
 func (u *userOperation) storeOperatorUserCredentials(user *nexus.User) error {
 	secret := &corev1.Secret{}
 	log.Debug("Attempt to store operator user credentials into Secret")
-	if err := framework.Fetch(u.k8sclient, framework.Key(u.nexus), secret); err != nil {
+	if err := framework.Fetch(u.k8sclient, framework.Key(u.nexus), secret, "Secret"); err != nil {
```
ricardozanini, line 109:
```suggestion
	if err := framework.Fetch(u.k8sclient, framework.Key(u.nexus), secret, framework.SecretKind); err != nil {
```

ricardozanini (approved):
LGTM! Minor comments only :)

pkg/controller/nexus/resource/networking/manager.go
LCaparelli: Ugh, not sure how that got in there. Thanks for pointing it out, definitely want to keep it :-)

pkg/controller/nexus/server/users.go
LCaparelli: Well spotted, thanks!

Kaitou786 (approved):

PR #172, opened 2020-10-06
Add goimports to hack/go-fmt.go

### LINKED ISSUE #171: Add goimports to hack/go-fmt.go
[goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) is meant to be a replacement for `gofmt`. Right now [it's not a perfect drop-in replacement](https://github.com/golang/go/issues/21476) for it, but we may still get the best of worlds by running both in `hack/go-fmt.sh`.

As `goimports` is not as common as `gofmt`, it's probably a good idea to make sure the binary is present, similarly to how we check if `openapi-gen` is installed and install it the same way it's done in [`hack/go-vet.sh`](https://github.com/m88i/nexus-operator/blob/main/hack/go-vet.sh#L25):

```
 24 # get the openapi binary
 25 command -v openapi-gen >/dev/null || go build -o $GOPATH/bin/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen
```

Kevin-Mok:
This fixes #171 in the first commit and runs `goimports` in the second.


LCaparelli (approved):
@Kevin-Mok thank you very much for your contribution! Looking good :-)

Issue #174, opened 2020-10-09
Use Admission Webhooks to set default values and validate to Nexus CR


ricardozanini:
We might need to get rid of the `validation` package or call it via Admission Webhooks after #140 

See: https://sdk.operatorframework.io/docs/building-operators/golang/webhooks/

LCaparelli:
This has some unforeseen complications.  The current state of implementation reworks a lot of our design in order to move the admission logic into the `v1alpha1` package (see [this issue](https://github.com/operator-framework/operator-sdk/issues/4249)).

In addition to that, mutating and validating webhooks require TLS termination, which in turn requires us to serve a TLS certificate. [Kubebuilder's approach](https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html) suggests the use of [cert-manager](https://cert-manager.io/) for injecting the CA bundle on the webhook configuration resources as well as for populating a secret containing a self-signed cert (and its private key).

This approach forces users of our Operator to  install cert-manager in order to use it, which is something we'd like to avoid.

Worth considering is that many Operators rely on the operator-sdk or kubebuilder at some level, and pulling cert-manager as a dependency might eventually become commonplace, in the sense that all operator users will already have it installed for one of the many operators installed which would require it.

In any case, some alternatives to avoid this dependency:

  1. not implementing validation through webhooks: simply go on as we do today, which consists of setting defaults and validating the CR on the reconcile loop
  2.  change kubebuilder-generated manifests to remove the volume which mounts the cert secret and generate the cert during initialization, populating  /tmp/k8s-webhook-server/serving-certs with it.
  3. create a sidecar container which would generate a self-signed cert and populate the webhook-server-cert secret with it, allowing it to be mounted on the manager container.

In the light of these complications, we should take a step back to analyze our options and further discuss the strategy moving forward. Moving this to v0.6.0 instead of v0.5.0. 

### My 2c
#### 1. not implementing validation through webhooks

I personally dislike this approach. It allows users to persist invalid CRs to the etcd store, in case of validation failures the user does not get immediately notified with an error message thrown to the terminal (they need to check the status of the CR or its logs to understand what is going on). It does not follow the suite of the core API resources. It changes the CR during the reconcile loop, which seems like anti-pattern.

#### 2. change kubebuilder-generated manifests and write to the container FS directly

[...]

PR #175, opened 2020-10-09
Closes #140 - Upgrading to kubebuilder format and SDK 1.0.1

### LINKED ISSUE #140: Upgrade to Operator SDK 1.0.0 and new Manifest Bundle
OpensShift 4.6 requires a new manifest bundle for operators and upgrading to SDK 1.0.0 will help with this change.

ricardozanini:
Closes #140 

TODO:

- [x] - Update documentation
- [x] - Migrate e2e tests
- [x] - Migrate hack scripts
- [x] - Migrate CI

:sweat: 


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
LCaparelli, line 1: shouldn't `.vscode` be in `.gitignore`?

bundle/manifests/nexus-operator.clusterserviceversion.yaml
```diff
@@ -32,69 +32,37 @@ metadata:
           }
```
LCaparelli, line 32: Should this file have been changed at all? It's 0.2

ricardozanini:
@LCaparelli There's a lot of clutter in this PR. It's not ready yet. I forgot to change to draft, don't even waste your time reviewing it 😅😅

bundle/manifests/nexus-operator.clusterserviceversion.yaml
ricardozanini: This file does not exist any more, GH is accusing a rename, but was deleted actually.

ricardozanini:
@LCaparelli @Kaitou786 @Kevin-Mok we're ready, I believe. If you guys have the time, please could you take a look? I'll get back at the end of the day to apply your suggestions, if any. 

There's a lot to do, but I believe I covered the vast majority in this PR. We can keep improving. Since 0.4.0 will be released at the end of the month, I think we will have the time to find any bugs that might come up duo to the migration.

bundle.Dockerfile
```diff
@@ -3,9 +3,15 @@ FROM scratch
...
+LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
+LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v2
+LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
+LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
 
-COPY deploy/olm-catalog/nexus-operator/manifests /manifests/
-COPY deploy/olm-catalog/nexus-operator/metadata/annotations.yaml /metadata/annotations.yaml
+COPY bundle/manifests /manifests/
+COPY bundle/metadata /metadata/
+COPY bundle/tests/scorecard /tests/scorecard/
```
Kaitou786, line 17: Just curious, where is this Dockerfile going to be used?

Kaitou786 (approved):
They changed **a lot** but this is nice <3

bundle.Dockerfile
Kaitou786: Ah nevermind, saw it

pkg/framework/fetcher.go
```diff
@@ -21,16 +21,13 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/m88i/nexus-operator/pkg/logger"
 )
 
 func Fetch(client client.Client, key types.NamespacedName, instance resource.KubernetesResource, kind string) error {
-	log := logger.GetLoggerWithNamespacedName("resource_fetcher", key)
```
LCaparelli, line 29: Why remove the instantiation of a new logger here? The idea is to already have the resource name as a field in the log.

pkg/framework/fetcher.go
```diff
@@ -21,16 +21,13 @@ import (
...
-	log.Debugf("Attempting to fetch deployed %s", kind)
+	log.Info("Attempting to fetch", "deployed", kind)
```
LCaparelli, line 27: IMO this is also a bit weird.

`{"msg":"Attempting to fetch","deployed":"Service"}"

I was thinking of:

```suggestion
	log.Info("Attempting to fetch deployed resource", "kind", kind)
```

Producing:

`{"msg":"Attempting to fetch deployed resource","kind":"Service"}"

Wdyt?

pkg/framework/fetcher.go
```diff
@@ -21,16 +21,13 @@ import (
...
 	if err := client.Get(ctx.TODO(), key, instance); err != nil {
 		if errors.IsNotFound(err) {
-			log.Debugf("There is no deployed %s", kind)
+			log.Debug("There is no deployed", "kind", kind)
```
LCaparelli, line 30: Here I think the previous format is a bit more straightforward, but I don't feel strongly about it. Up to you

pkg/test/client.go
```diff
@@ -17,7 +17,7 @@ package test
 import (
 	"context"
 
-	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
+	openapi_v2 "github.com/googleapis/gnostic/openapiv2"
```
LCaparelli, line 20: :tada:

LCaparelli (commented):
Thanks a lot for this @ricardozanini, looks like it was a handful. Also sorry for the delay.

Left some comments in

Makefile
```diff
@@ -1,76 +1,148 @@
...
-
-.PHONY: uninstall
-uninstall:
-	./hack/uninstall.sh
-
-.PHONY: olm-integration
-olm-integration:
-	./hack/olm-integration.sh
-
-.PHONY: test-e2e
```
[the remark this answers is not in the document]
ricardozanini: The e2e was replaced by the [envtest](https://sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/#migrate-your-tests) approach, that's why I removed these tests completely.

Makefile
```diff
@@ -1,76 +1,148 @@
...
-namespace="nexus-e2e"
-create_namespace=true
-test-e2e:
-	NAMESPACE_E2E=$(namespace) CREATE_NAMESPACE=$(create_namespace) ./hack/run-e2e-test.sh
-
-.PHONY: pr-prep
-create_namespace=true
-run_with_image=true
-pr-prep:
-	CREATE_NAMESPACE=$(create_namespace) RUN_WITH_IMAGE=$(run_with_image) ./hack/pr-prep.sh
```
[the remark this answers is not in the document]
ricardozanini: oh sorry, I'll add it back.

controllers/nexus/resource/networking/manager.go
```diff
@@ -99,7 +97,7 @@ func (m *Manager) GetRequiredResources() ([]resource.KubernetesResource, error)
 			return nil, fmt.Errorf(resUnavailableFormat, "routes")
 		}
 
-		m.log.Debugf("Generating required %s", framework.RouteKind)
+		m.log.Debug("Generating required", "resources", framework.RouteKind)
```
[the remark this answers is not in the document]
ricardozanini: gimme a break, too many logs messages to review. I deprecated the `Debugf` method, but then golanglint started to scream at me. :D

controllers/nexus/resource/validation/validation.go
```diff
@@ -84,37 +82,37 @@ func (v *Validator) validate(nexus *v1alpha1.Nexus) error {
...
 func (v *Validator) validateNetworking(nexus *v1alpha1.Nexus) error {
 	if !nexus.Spec.Networking.Expose {
-		v.log.Debugf("'spec.networking.expose' set to 'false', ignoring networking configuration")
+		v.log.Debug("'spec.networking.expose' set to 'false', ignoring networking configuration")
 		return nil
 	}
 
 	if !v.ingressAvailable && nexus.Spec.Networking.ExposeAs == v1alpha1.IngressExposeType {
-		v.log.Errorf("Ingresses are not available on your cluster. Make sure to be running Kubernetes > 1.14 or if you're running Openshift set 'spec.networking.exposeAs' to '%s'. Alternatively you may also try '%s'", v1alpha1.IngressExposeType, v1alpha1.NodePortExposeType)
+		v.log.Info("Ingresses are not available on your cluster. Make sure to be running Kubernetes > 1.14 or if you're running Openshift set ", "spec.networking.exposeAs", v1alpha1.IngressExposeType, "Also try", v1alpha1.NodePortExposeType)
```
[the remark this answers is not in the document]
ricardozanini: Those are not errors since we don't have an error instance. We can consider review the validation approach (#174), or just raise a `errors.New()` here.

controllers/nexus/resource/validation/validation.go
```diff
@@ -204,43 +202,43 @@ func (v *Validator) setUpdateDefaults(nexus *v1alpha1.Nexus) {
...
 		return
 	}
 
 	if nexus.Spec.AutomaticUpdate.MinorVersion == nil {
-		v.log.Debugf("Automatic Updates are enabled, but no minor was informed. Fetching the most recent...")
+		v.log.Debug("Automatic Updates are enabled, but no minor was informed. Fetching the most recent...")
 		minor, err := update.GetLatestMinor()
 		if err != nil {
-			v.log.Errorf("Unable to fetch the most recent minor: %v. Disabling automatic updates.", err)
+			v.log.Error(err, "Unable to fetch the most recent minor: %v. Disabling automatic updates.")
```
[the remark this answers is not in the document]
ricardozanini: No need. The `Error` method interface is `Error(error, string, keyAndValues...)`

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
LCaparelli: Ah, I see. #174 looks great, but in the meantime I think we should go for `errors.New()` here with this message.

controllers/nexus/update/tags.go
```diff
@@ -123,7 +123,7 @@ func getTags() ([]string, error) {
 	// redirect the lib's logging to ours
 	reg.Logf = func(format string, args ...interface{}) {
 		format = fmt.Sprintf("Registry: %s", format)
-		log.Infof(format, args)
+		log.Info(format, args...)
```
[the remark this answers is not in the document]
ricardozanini: I don't know, I have to see it in action. Do we have a unit test covering this use case?

pkg/framework/fetcher.go
ricardozanini: we already have a log instance in this package, no need to create a new one. :)

pkg/framework/fetcher.go
ricardozanini: We could rephrase this sentence. Any ideas?

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
LCaparelli: Right, but the `msg string` [parameter](https://github.com/go-logr/logr/blob/master/logr.go#L169) doesn't seem to be a format, won't the "`%v`" be printed?

In [klog's implementation](https://github.com/kubernetes/klog/blob/master/klogr/klogr.go#L164) this parameter is handled as a simple string (not as a format) and if I'm reading this correctly, the `%v` would be printed normally.

pkg/framework/fetcher.go
LCaparelli: Right, but that instance doesn't identify the name of the resource being fetched. I'm ok with not instantiating a new one if we change the actual below into something like:

`log.Info("Attempting to fetch deployed resource", "kind", kind, "namespacedName", key)`

Wdyt?

pkg/framework/fetcher.go
LCaparelli: Hum, maybe:

```suggestion
			log.Debug("Unable to find resource", "kind", kind, "namespacedName", key)
```

Wdyt?

controllers/nexus/resource/networking/manager.go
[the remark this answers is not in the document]
LCaparelli: Yeah, lots of messages, sorry :-( 

We need to have consistent logging so it can be easily parsed with greps, for example. If a single log message is different from other points in the code it wouldn't be easily identified when troubleshooting.

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
ricardozanini: I believe here we can go either with `errors.New()` or `log.Warn()`, I'll take a closer look to it.

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
ricardozanini: Oh noes.  I need to remove the `%v` format. Just that.

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
LCaparelli: I believe error would be better, this leads to a reconcile error. When we have validation webhooks this will even prevent the user from storing the resource altogether. A warning implies the standard flow will be followed (though possibly with odd behavior), which isn't the case.

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
LCaparelli: Yeah, that's what I meant hahahaha sorry, should have made a suggestion so it's clear.

```suggestion
			v.log.Error(err, "Unable to fetch the most recent minor. Disabling automatic updates.")
```

controllers/nexus/resource/validation/validation.go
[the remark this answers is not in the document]
ricardozanini: Changed o `Warn` since in the next line we do raise the actual error.

controllers/nexus/update/tags.go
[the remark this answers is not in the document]
ricardozanini: I just called format again to favor their formatting behavior and then call our log function to display it.

controllers/nexus/update/tags.go
```diff
@@ -122,8 +122,8 @@ func getTags() ([]string, error) {
 	}
 	// redirect the lib's logging to ours
 	reg.Logf = func(format string, args ...interface{}) {
-		format = fmt.Sprintf("Registry: %s", format)
-		log.Info(format, args...)
+		format = fmt.Sprintf(fmt.Sprintf("Registry: %s", format), args)
```
LCaparelli, line 125: Nice!

controllers/nexus/resource/validation/validation.go
```diff
@@ -87,32 +87,32 @@ func (v *Validator) validateNetworking(nexus *v1alpha1.Nexus) error {
 	}
 
 	if !v.ingressAvailable && nexus.Spec.Networking.ExposeAs == v1alpha1.IngressExposeType {
-		v.log.Info("Ingresses are not available on your cluster. Make sure to be running Kubernetes > 1.14 or if you're running Openshift set ", "spec.networking.exposeAs", v1alpha1.IngressExposeType, "Also try", v1alpha1.NodePortExposeType)
+		v.log.Warn("Ingresses are not available on your cluster. Make sure to be running Kubernetes > 1.14 or if you're running Openshift set ", "spec.networking.exposeAs", v1alpha1.IngressExposeType, "Also try", v1alpha1.NodePortExposeType)
```
LCaparelli, line 90: So you went with warnings, huh? Up to you, but I still think we should log it as error. Shouldn't cause too much trouble after #174 anyways

ricardozanini:
@LCaparelli I believe I covered all your comments, can you take a look again please?

LCaparelli (approved):
Thanks again @ricardozanini :-)

controllers/nexus/resource/validation/validation.go
ricardozanini: Take a look in the next line, you will see that there's an error being raised.

PR #177, opened 2020-10-18
Closes #149 - Add Maven public URL to the Nexus CR status

### LINKED ISSUE #149: Add Maven Public repository URL to the Nexus CR Status
Add Maven Public repository URL (internal and external) to the Nexus CR Status

This URL can be grabbed from the deployed Nexus instance after deploying it. The internal URL should be pretty straightforward since it's basically the Kubernetes Service deployed by the operator. Refer to the Kubernetes documentation for more information: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

The Public URL will vary according to the `exposeAs` attribute in the CR. `NodePort` is way simpler, but we will need to figure the external IP for the cluster. This can be trick, and might need a new Issue to implement this feature. 

Users can use the `exposeAs` to automatically make the operator to create an Ingress (Kubernetes) or Route (OpenShift). So the easy path would be querying these resources after being deployed and add the URL to the Nexus CR.

But if users decide to expose the URL themselves, one option would be to query the Services linked to the Deployment and create the URL from there.

When running `kubectl get nexus/nexus3`, I'd like to see the maven public group URL displayed in the console, this way I wouldn't have to login to the Nexus UI to grab the URL and add to my objects.

This URL can be the one exposed internally by the Kubernetes cluster, since public (exposed) URLs might be trick to accomplish since varies a lot. A new issue might follow this one to implement such feature.

Login to the Nexus UI and grab the URL there

**Additional context**
Would help users to configure their apps faster

ricardozanini:

Closes #149 

I need to run some smoke tests to make sure that this is working properly on k8s.

### DIFF
--- a/api/v1alpha1/nexus_types.go
+++ b/api/v1alpha1/nexus_types.go
@@ -249,6 +249,7 @@ type OperationsStatus struct {
 	CommunityRepositoriesCreated bool   `json:"communityRepositoriesCreated,omitempty"`
 	MavenCentralUpdated          bool   `json:"mavenCentralUpdated,omitempty"`
 	Reason                       string `json:"reason,omitempty"`
+	MavenPublicURL               string `json:"mavenPublicURL,omitempty"`
 }
 
 type NexusStatusType string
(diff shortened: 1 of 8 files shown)


ricardozanini:
It's alive!

```
oc get nexus
NAME      EXPOSE AS   UPDATE DISABLED   STATUS    REASON    MAVEN PUBLIC URL
nexus3    NodePort                      OK                  http://192.168.39.209:31031/repository/maven-public/
```

Kaitou786 (approved):
Looks good

PR #181, opened 2020-10-19
Move discovery to a separate package

### LINKED ISSUE #178: Move discovery to a separate package
Currently discovery feels a bit awkward to use within the controller logic as each function must carry and pass around the `DiscoveryInterface` interface as argument whenever a call further down needs it, such as when resource managers need discovery:

https://github.com/m88i/nexus-operator/blob/8bcda688a3039c1672b4cf07e6e0ff7a97a20484/controllers/nexus/resource/resources.go#L55-L58

We could avoid this by moving discovery to its own package, which would receive the `DiscoveryInterface` during initialization and other packages would have discovery via this package without the need for passing client interfaces around business logic code.

This would move code from `pkg/cluster/openshift` and `pkg/cluster/kubernetes` into a separate package (e.g., `pkg/cluster/discovery`).

LCaparelli:
Fix #178

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/m88i/nexus-operator:$(VERSION)
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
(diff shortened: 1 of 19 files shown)


LCaparelli:
Huh, unit tests failing on the CI but working fine locally when I run `make test`. I'll see what I can dig up

Makefile
```diff
@@ -13,7 +13,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/m88i/nexus-operator:$(VERSION)
```
ricardozanini, line 16: Not sure if we want to change the SDK generated code. I added `OPERATOR_IMG` for this purpose. Have you researched about this? Can we change it? Not sure yet.

If you feel comfortable to change it, please remove `OPERATOR_IMG` variable.

ricardozanini (changes requested):

Makefile
LCaparelli: Ah my bad, I changed it because `make docker-build` would build with this weird label and `make deploy` wouldn't work properly, but I just saw there is the `operator-build` recipe which does more or less the same and with the behavior I was expecting. I'll change this back, thanks!

ricardozanini (approved):

Kaitou786 (approved):

PR #182, opened 2020-10-28
Use v1 Ingresses when available 

### LINKED ISSUE #180: When available, use networking.k8s.io/v1 API version for Ingresses
`networking.k8s.io/v1beta1` is deprecated in Kubernetes v1.19+ and running the operator in more recent clusters runs into:

`W1018 19:47:08.954666   14552 warnings.go:67] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress` 

It would be nice if the Operator used `v1` Ingresses when available.

N/A

**Additional context**
N/A

Yes

LCaparelli:
Fix #180

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,12 +34,14 @@ all: manager
 
 # Run tests
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
+# Needed to support k8s.io/api/networking/v1 Ingress
+K8S_VERSION=1.19.0
 test: generate-installer fmt vet bundle
 	mkdir -p ${ENVTEST_ASSETS_DIR}
 	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
 	sed -i "s,#\!.*,#\!\/bin\/bash,g" ${ENVTEST_ASSETS_DIR}/setup-envtest.sh
 	sed -i "/pipefail/d" ${ENVTEST_ASSETS_DIR}/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
+	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; ENVTEST_K8S_VERSION=$(K8S_VERSION) fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
 
 generate-installer: generate manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(OPERATOR_IMG)
--- a/controllers/nexus/resource/deployment/manager.go
+++ b/controllers/nexus/resource/deployment/manager.go
@@ -28,12 +28,13 @@ import (
 
 	"github.com/m88i/nexus-operator/api/v1alpha1"
 	"github.com/m88i/nexus-operator/pkg/framework"
+	"github.com/m88i/nexus-operator/pkg/framework/kind"
 	"github.com/m88i/nexus-operator/pkg/logger"
 )
 
 var managedObjectsRef = map[string]resource.KubernetesResource{
-	framework.DeploymentKind: &appsv1.Deployment{},
-	framework.ServiceKind:    &corev1.Service{},
+	kind.DeploymentKind: &appsv1.Deployment{},
+	kind.ServiceKind:    &corev1.Service{},
 }
 
 // Manager is responsible for creating deployment-related resources, fetching deployed ones and comparing them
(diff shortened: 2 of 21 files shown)


LCaparelli:
Actually, I just realized in the shower there's a bug in this. It's not enough to match Group and Version (as we have been doing so far). We must match Group, Version and Kind. In previous versions of K8s, such as 1.16, has `k8s.io/api/networking/v1` but Ingress is not one of the available Kinds in this version.

ricardozanini (changes requested):
A few comments about the desing :D 
Tell me what you think. I believe we can stretch the release a little bit to include this PR on 0.4.0.

Makefile
```diff
@@ -34,12 +34,14 @@ all: manager
...
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
+# Needed to support k8s.io/api/networking/v1 Ingress
+K8S_VERSION=1.19.0
 test: generate-installer fmt vet bundle
 	mkdir -p ${ENVTEST_ASSETS_DIR}
 	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
 	sed -i "s,#\!.*,#\!\/bin\/bash,g" ${ENVTEST_ASSETS_DIR}/setup-envtest.sh
 	sed -i "/pipefail/d" ${ENVTEST_ASSETS_DIR}/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
+	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; ENVTEST_K8S_VERSION=$(K8S_VERSION) fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
```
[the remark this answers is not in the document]
LCaparelli: Yup :-)

controllers/nexus/resource/networking/manager.go
```diff
@@ -32,23 +32,36 @@ import (
...
-	nexus                                 *v1alpha1.Nexus
-	client                                client.Client
-	log                                   logger.Logger
-	routeAvailable, ingressAvailable, ocp bool
+	nexus             *v1alpha1.Nexus
+	client            client.Client
+	log               logger.Logger
+	managedObjectsRef map[string]resource.KubernetesResource
+
+	routeAvailable, ingressAvailable, legacyIngressAvailable bool
```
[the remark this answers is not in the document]
LCaparelli: That's a great idea! Thanks! I'll see what I can come up with

LCaparelli:
Thanks for reviewing @ricardozanini, I just pushed another commit fixing discovery in case you want to check it out. I'll work on the suggestion :-)

controllers/nexus/resource/deployment/manager.go
```diff
@@ -28,12 +28,13 @@ import (
...
 	"github.com/m88i/nexus-operator/pkg/framework"
+	"github.com/m88i/nexus-operator/pkg/framework/kind"
 	"github.com/m88i/nexus-operator/pkg/logger"
 )
 
 var managedObjectsRef = map[string]resource.KubernetesResource{
-	framework.DeploymentKind: &appsv1.Deployment{},
-	framework.ServiceKind:    &corev1.Service{},
+	kind.DeploymentKind: &appsv1.Deployment{},
+	kind.ServiceKind:    &corev1.Service{},
```
ricardozanini, line 37: I would avoid creating too many packages. What's the problem with `framework` for common k8s API features?

controllers/nexus/resource/networking/manager.go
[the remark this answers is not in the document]
LCaparelli: @ricardozanini I've been trying to come up with an implementation where this mechanism actually eliminates complexity, but all ideas I had so far actually turned out to be more complicated than boolean flags. The recurring theme in my attempts is that we'll need to perform this checks one way or another, so this sort of structure ends up further complicating logic.

These experiments made me realize we're also duplicating work here from the validation package and also made me realize some problems in how we validate things.

I think a rework of this design is in order, but perhaps we should wait for the implementation of validation webhooks and then we'll have a clearer picture of how we can work this out.

Wdyt of keeping things as they are for now and we'll rethink the design in a different issue once we're done migrating the validation?

controllers/nexus/resource/deployment/manager.go
LCaparelli: Needed that to avoid an import cycle: the fake client now needs the Kinds in order to populate the API Resources list from each GroupVersion, but the tests from `framework` actually use the fake client. So these had to be moved to a separate package.

But why do you want to avoid having too many packages?

ricardozanini: > But why do you want to avoid having too many packages?

[Keep this in mind](https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_consider_fewer_larger_packages).

controllers/nexus/resource/networking/manager.go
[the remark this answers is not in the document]
ricardozanini: Agreed to postpone this refactor to #183 and #174

controllers/nexus/resource/deployment/manager.go
LCaparelli: That's reasonable. What if we merge the `test` package into `framework`?

ricardozanini: the `test` package shouldn't be included in the final binary, so it's reasonable to keep it there. That's ok having `kind`, but I'll keep an eye on new packages creation, just to make sure we are not abusing.

LCaparelli:
@ricardozanini any other suggestions, concerns or nitpicks? :-)

ricardozanini (approved):

ricardozanini:
I believe we can merge by the end of the day. Tomorrow morning I'll update the image and the manifest, so we can run the tests during the weekend to release 0.4.0.

Kaitou786 (approved):
This is nice!

PR #192, opened 2020-12-09
Fix #191 - Changing deployment strategy to recreate and setting replicas to max 1

### LINKED ISSUE #191: Pod fails to start after modifying the Nexus resource
**Describe the bug**
Modifying the Nexus resource triggers a new Deployment, but the new Pod can't start (CrashLoopBackOff) because the previous one is still holding the /nexus-data/lock. The problem is probably cause by the Deployment using spec.strategy=RollingUpdate. Using "Recreate" may help, so previous Nexus instance is shut down before the new one is created.

**To Reproduce**
1. Create a Nexus resource
2. Modify the Nexus resource (e.g. mem requests)
3. The new Pod won't start, but keeps crashing.

**Expected behavior**
The Deployment is successfully rolled out.

**Environment**
OpenShift 4.6.5
Client Version: 4.4.30
Server Version: 4.6.5
Kubernetes Version: v1.19.0+9f84db3

**Additional context**

ricardozanini:
Fixes #191 

@LCaparelli do you mind taking a look? Maybe improving the doc a little bit? I'm lacking creativity...

@bszeti I'll push an image with this fix soon, do you mind taking testing in your local environment? As soon as we review and test this PR, I'll release a new version with this fix for you.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Table of Contents
       * [Red Hat Certified Images](#red-hat-certified-images)
       * [Image Pull Policy](#image-pull-policy)
       * [Repositories Auto Creation](#repositories-auto-creation)
+      * [Scaling](#scaling)
       * [Contributing](#contributing)
 
 
@@ -412,6 +413,37 @@ The Operator also will create three Maven repositories by default:
 All of these repositories will be also added to the `maven-public` group. This group will gather the vast majority of jars needed by the most common use cases out there. If you won't need them, just disable this behavior by setting the attribute `spec.serverOperatons.disableRepositoryCreation` to `true` in the Nexus CR. 
 
 All of these operations are disabled if the attribute `spec.generateRandomAdminPassword` is set to `true`, since default credentials are needed to create the `nexus-operator` user. You can safely change the default credentials after this user has been created.
+
+## Scaling
+
+For now, the Nexus Operator won't accept a number higher than `1` to the `spec.replicas` attribute.
+This is because the Nexus server can't share its mounted persistent volume with other pods. See #191 for more details.
+
+Horizontal scaling will only work once we add HA support to the operator (see #61). 
+If you need to scale the server, you should take the vertical approach and increase the numbers of resource limits used
+by the Nexus server. For example:
+
+```yaml
+apiVersion: apps.m88i.io/v1alpha1
+kind: Nexus
+metadata:
+  name: nexus3
+spec:
+  replicas: 1
+  # Set the resources requests and limits for Nexus pods. See: https://help.sonatype.com/repomanager3/system-requirements
+  resources:
+    limits:
+      cpu: "4"
+      memory: "8Gi"
+    requests:
+      cpu: "1"
+      memory: "2Gi"
+  persistence:
+    persistent: true
+    volumeSize: 10Gi
+```
+
+We are working to support HA in the future.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ All of these operations are disabled if the attribute `spec.generateRandomAdminP
 For now, the Nexus Operator won't accept a number higher than `1` to the `spec.replicas` attribute.
 This is because the Nexus server can't share its mounted persistent volume with other pods. See #191 for more details.
 
-Horizontal scaling will only work once we add HA support to the operator (see #61). 
+Horizontal scaling will only work once we add [HA support](https://help.sonatype.com/repomanager3/high-availability) to the operator (see #61). 
 If you need to scale the server, you should take the vertical approach and increase the numbers of resource limits used
 by the Nexus server. For example:
 
--- a/controllers/nexus/resource/validation/validation.go
+++ b/controllers/nexus/resource/validation/validation.go
@@ -80,7 +80,21 @@ func (v *Validator) SetDefaultsAndValidate(nexus *v1alpha1.Nexus) (*v1alpha1.Nex
 }
 
 func (v *Validator) validate(nexus *v1alpha1.Nexus) error {
-	return v.validateNetworking(nexus)
+	validators := []func(*v1alpha1.Nexus) error{v.validateDeployment, v.validateNetworking}
+	for _, v := range validators {
+		if err := v(nexus); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (v *Validator) validateDeployment(nexus *v1alpha1.Nexus) error {
+	if nexus.Spec.Replicas > 1 {
+		v.log.Warn("Nexus server only supports 1 replica.", "Desired Replicas", nexus.Spec.Replicas)
+		nexus.Spec.Replicas = ensureMaximum(nexus.Spec.Replicas, 1)
+	}
+	return nil
 }
 
 func (v *Validator) validateNetworking(nexus *v1alpha1.Nexus) error {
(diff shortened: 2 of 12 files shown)


ricardozanini:
@LCaparelli @bszeti the image is available at: `quay.io/m88i/nexus-operator:0.5.0-cr1` can be verified with the `[nexus-operator.yaml` file](https://github.com/m88i/nexus-operator/blob/main/nexus-operator.yaml). Just change the image tag there.

ricardozanini:
OLM test may fail since they changed some bits. @Kaitou786 since you've changed that on Kogito, mind replicating the fix here as well? :hand_over_mouth:

README.md
```diff
@@ -413,6 +414,37 @@ All of these repositories will be also added to the `maven-public` group. This g
 
 All of these operations are disabled if the attribute `spec.generateRandomAdminPassword` is set to `true`, since default credentials are needed to create the `nexus-operator` user. You can safely change the default credentials after this user has been created.
 
+## Scaling
+
+For now, the Nexus Operator won't accept a number higher than `1` to the `spec.replicas` attribute.
+This is because the Nexus server can't share its mounted persistent volume with other pods. See #191 for more details.
+
+Horizontal scaling will only work once we add HA support to the operator (see #61).
```
LCaparelli, line 422:
```suggestion
Horizontal scaling will only work once we add [HA support](https://help.sonatype.com/repomanager3/high-availability) to the operator (see #61). 
```

controllers/nexus/resource/validation/validation.go
```diff
@@ -80,7 +80,21 @@ func (v *Validator) SetDefaultsAndValidate(nexus *v1alpha1.Nexus) (*v1alpha1.Nex
...
+		}
+	}
+	return nil
+}
+
+func (v *Validator) validateDeployment(nexus *v1alpha1.Nexus) error {
+	if nexus.Spec.Replicas > 1 {
+		v.log.Warn("Nexus server only supports 1 replica.", "Desired Replicas", nexus.Spec.Replicas)
+		nexus.Spec.Replicas = ensureMaximum(nexus.Spec.Replicas, 1)
+	}
```
LCaparelli, line 96: This should be in the default setting logic IMO. Validation shouldn't modify the resource. This is also aligned with how admission webhooks are implemented, where there is a explicit separation between the mutating webhook (responsible for default-setting and any other changes made to the CR) and the validating webhook (makes no modification and simply returns an error if needed).

Wdyt?

LCaparelli (changes requested):
Thanks a lot for this @ricardozanini! I didn't find the doc section lacking, it's straightforward and clear, nothing to add on my behalf but linking the HA page from Sonatype.

I do think the logic that controls `spec.replica` should be placed at a default setting method, not a validation one.

controllers/nexus/resource/validation/validation.go
```diff
@@ -80,7 +80,21 @@ func (v *Validator) SetDefaultsAndValidate(nexus *v1alpha1.Nexus) (*v1alpha1.Nex
...
+		if err := v(nexus); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (v *Validator) validateDeployment(nexus *v1alpha1.Nexus) error {
+	if nexus.Spec.Replicas > 1 {
+		v.log.Warn("Nexus server only supports 1 replica.", "Desired Replicas", nexus.Spec.Replicas)
```
LCaparelli, line 94: Also, we should avoid whitespace in log keys to keep consistent with the rest of the project.

```suggestion
		v.log.Warn("Nexus server only supports 1 replica.", "DesiredReplicas", nexus.Spec.Replicas)
```

Wdyt?

LCaparelli:
@ricardozanini just tested the release candidate and things are working as expected. Setting `spec.replicas` to 3 changes it back to 1, it gets logged fine and any changes which cause a new ReplicaSet to be created now destroy the existing pods from the previous RS first. :-D

controllers/nexus/resource/validation/validation.go
ricardozanini: Yeah, I agree. IDK why I've added in there. I'll chang it.

LCaparelli: I agree it's not a default, but it's a mutation of the CR. That's the main reason why I don't believe we should keep it in validation. The way it is, cohesion is reduced as the validation method is validating AND changing state. IMO, we should first make all mutations to the state and only then validate.

This is already how we ensure values for the probes:

https://github.com/m88i/nexus-operator/blob/126e9042352ffe299e41b3a0951ce1f1d5fb8708/controllers/nexus/resource/validation/validation.go#L167-L198

controllers/nexus/resource/validation/validation.go
LCaparelli: Looks like this one didn't make it into the changes @ricardozanini. Did you forget it or do you not agree? hehe :-)

controllers/nexus/resource/validation/validation.go
ricardozanini: If I change the mutation to defaults, defaults will be doing validation since I have to check if it's greater than 1 or not. There are no defaults. :)

So there's no cohesion also moving to set defaults.

controllers/nexus/resource/validation/validation.go
ricardozanini: LOL

controllers/nexus/resource/validation/validation.go
ricardozanini: > we should first make all mutations to the state and only then validate.

This does not make sense in this scenario. We call the default first, which will have to validate the field, if it's false we mutate to a valid one. So are you suggesting to validate and mutate in default?

LCaparelli: When I think of validation, I think of simply checking the values and returning an error if they're not valid.

I think the main issue here is thinking of the phase where we change the CR as default-setting only. It's not. It's "make any changes necessary prior validation". I believe we have more to gain from making this separation than we have from altering state during validation.

More appropriate names for each procedure would be `mutate` and `validate`, following the webhook definitions.

["After all object modifications are complete, and after the incoming object is validated by the API server, validating admission webhooks are invoked and can reject requests to enforce custom policies."](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks)

When a validation webhook is called, its response is either ok or not. You can't make changes during that, they won't be persisted by the API server. The only response which can modify incoming objects can only come from mutating webhooks:

["When allowing a request, a mutating admission webhook may optionally modify the incoming object as well. This is done using the patch and patchType fields in the response. The only currently supported patchType is JSONPatch. See JSON patch documentation for more details. For patchType: JSONPatch, the patch field contains a base64-encoded array of JSON patch operations."](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response)

Even if we keep this in validation now, it will need to be moved to mutation in #174.

ricardozanini: +1 to renaming. I'll move to the `defaults` and let your PR handle the refactoring with the Admission controllers.

ricardozanini:
@LCaparelli can you take a final look? I've moved to `defaults` instead. Hopefully, with admission controllers, we wouldn't need that, since the CR won't be created.

LCaparelli (approved):

LCaparelli:
> @LCaparelli can you take a final look? I've moved to `defaults` instead.

This looks great man, thanks! 

> Hopefully, with admission controllers, we wouldn't need that, since the CR won't be created.

Well, we will, but it will be part of the mutating webhook, not part of the reconciliation logic.

PR #221, opened 2022-01-05
Fix apiVersion not supported in new k8s

radtriste:


ricardozanini (approved):

ricardozanini:
> not sure what is the problem there ...

I need to update the OLM tests

